PR #17, opened 2026-01-12
RFC: Dynamic webhook middleware for ToolHive

JAORMX:
## Summary

Port proposal from [toolhive PR #2300](https://github.com/stacklok/toolhive/pull/2300).

Introduces a dynamic webhook middleware system for ToolHive that enables external services to participate in the MCP request pipeline, inspired by Kubernetes admission webhooks.

## Problem

Organizations need to integrate ToolHive with existing enterprise systems (policy engines, SIEM, ticketing, HR, CMDB) but currently require forking ToolHive and maintaining custom builds to add new middleware functionality.

## Solution

Two new webhook types:
- **Validating webhooks**: Accept/deny requests based on external policy decisions
- **Mutating webhooks**: Transform requests (enrich with context, redact PII, etc.)

## Key Features

- Multiple webhooks with sequential execution (pipeline pattern)
- File-based CLI configuration (`--webhook-config webhook.yaml`)
- Kubernetes MCPWebhookConfig CRD for operator
- TLS/mTLS support with file-based certificates
- Configurable failure policies (fail-open/fail-closed)
- Full observability (Prometheus metrics, audit logs, OpenTelemetry tracing)
- Versioned protocol (`version: v0.1.0`) for evolution

## Use Cases

1. **External Policy Engine Integration** - Use OPA or custom policy engines
2. **Request Enrichment** - Add context from HR/CMDB systems
3. **Approval Workflows** - Require manager approval for high-risk operations
4. **SIEM Integration** - Send custom-formatted logs to enterprise SIEM
5. **Rate Limiting** - Enforce limits using external rate limiters
6. **PII Redaction** - Remove sensitive data before processing

## Architecture

Webhooks integrate into existing middleware chain:
1. Authentication -> 2. Token Exchange -> 3. MCP Parser -> 4. Tool Filter -> **5. Mutating Webhooks -> 6. Validating Webhooks** -> 7. Telemetry -> 8. Authorization -> 9. Audit

Multiple webhooks execute sequentially in configuration order.

### DIFF
--- a/rfcs/THV-0017-dynamic-webhook-middleware.md
+++ b/rfcs/THV-0017-dynamic-webhook-middleware.md
@@ -127,6 +127,7 @@ sequenceDiagram
     }
   },
   "mcp_request": {
+    "mcp_version": "2024-11-05",
     "method": "tools/call",
     "resource_id": "database_query",
     "arguments": {
@@ -135,13 +136,21 @@ sequenceDiagram
     }
   },
   "context": {
-    "server_name": "postgres-mcp",
+    "server_name": "my-vmcp-instance",
+    "backend_server": "postgres-mcp",
     "namespace": "production",
     "source_ip": "192.0.2.1",
     "transport": "sse"
   }
 }
 ```
+
+> **Context Fields**:
+> - `server_name`: The ToolHive/vMCP instance name handling the request
+> - `backend_server`: (Optional) The actual MCP server being proxied, when using vMCP
+> - `namespace`: Kubernetes namespace (if applicable)
+> - `source_ip`: Client IP address
+> - `transport`: Connection transport type (sse, stdio, etc.)
 
 **Validating Webhook Response** (200 OK with JSON body):
 ```json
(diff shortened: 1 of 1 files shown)


rfcs/THV-0017-dynamic-webhook-middleware.md
```diff
@@ -0,0 +1,532 @@
...
+  namespace: mcp-servers
+spec:
+  type: validating  # or "mutating"
+  url: https://policy.company.com/validate
+  failure_policy: fail  # or "ignore"
+  timeout: 5s
+  client_cert_ref:
+    secret_name: webhook-mtls-cert
+    cert_key: tls.crt
+    key_key: tls.key
```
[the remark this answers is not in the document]
jhrozek: we defer the webhook implementation to later phases, but the way I was thinking about this is that a "compliant" webhook should verify signatures on the payload

rfcs/THV-0017-dynamic-webhook-middleware.md
```diff
@@ -0,0 +1,532 @@
...
+```json
+{
+  "version": "v0.1.0",
+  "uid": "unique-request-id-for-idempotency",
+  "allowed": true,
+  "patch_type": "json_patch",
+  "patch": [
+    {
+      "op": "add",
+      "path": "/params/arguments/audit_user",
```
jhrozek: I wonder if we should put the request data into its own container e.g. `mcp_request`, currently attributes like `methods` or `params` are top-level. The reason being is that the mutating webhook should /only/ be allowed to mutate the MCP request, not e.g. the principal or the claims :-) and having them inside a container and the mutating path relative to the container would make it harder to screw up the implementation.

rfcs/THV-0017-dynamic-webhook-middleware.md
```diff
@@ -0,0 +1,532 @@
...
+  "type": "webhook_invocation",
+  "logged_at": "2025-10-23T10:30:00Z",
+  "outcome": "denied",
+  "component": "toolhive-webhook",
+  "webhook": {
+    "name": "external-policy",
+    "type": "validating",
+    "url": "https://policy.company.com/validate",
+    "duration_ms": 150,
+    "status_code": 200
```
jhrozek: what about mutating webhooks? Would we also audit which webhook mutated what?

rfcs/THV-0017-dynamic-webhook-middleware.md
JAORMX: Love it - wrapped everything in mcp_request now. Patches are scoped to that container so webhooks can't mess with principal/context.

PR #20, opened 2026-01-12
RFC: CLI and Desktop App version alignment

samuv:
### Summary

Establishes how ToolHive Studio (Desktop) manages CLI installation to guarantee version compatibility.

### Problem

- Desktop UI is 1:1 coupled to CLI's `thv serve` API (currently in alpha)
- Users can install CLI independently (Homebrew, direct download)
- Version mismatches cause API errors and broken UI
- No coordination exists between CLI and Desktop installations

### Solution

**Desktop owns CLI installation** — regardless of how Desktop was installed:

- Bundles CLI binary inside the app (no network required)
- Installs to `~/.toolhive/bin` on first launch  
- Auto-configures PATH (shadows existing installations)
- Validates version on every launch
- Provides conflict dialog when external CLI detected

### Why This Is Needed (Even After Alpha)

This infrastructure is required regardless of API maturity:

1. **Now**: Exact version pinning (alpha API, no compatibility guarantees)
2. **Future**: Relax to minimum version constraints once backward-compatibility policy exists
3. **Always**: Users need a mechanism to get CLI when installing Desktop

### Learned from Docker Desktop

| Problem | Docker | ToolHive |
|---------|--------|----------|
| Naming | Same name causes conflicts | Distinct: `toolhive` vs `toolhive-studio` |
| PATH setup | Manual for "User" mode | Auto-configured |
| Coexistence | Hard failures | Shadows in PATH, no conflicts |
| Admin required | Yes for "System" mode | Never |

### Key Design Decisions

- User-space installation (no elevation required)
- Atomic upgrades with checksum verification  
- Distinct package names to avoid Homebrew conflicts
- Foundation for future API version negotiation

### DIFF
--- a/rfcs/THV-0020-desktop-cli-version-alignment.md
+++ b/rfcs/THV-0020-desktop-cli-version-alignment.md
@@ -98,7 +98,7 @@ flowchart TD
     B -->|Winget| C
     B -->|Any Linux PM| C
     
-    C --> D[First Launch]
+    C --> D[Desktop Startup]
     D --> E{Detect External CLI}
     
     E -->|External CLI found| H[HARD STOP: Uninstall required]


rfcs/THV-0020-desktop-cli-version-alignment.md
```diff
@@ -0,0 +1,854 @@
...
+
+**If version matches:** Ready, no action needed.
+
+**If version mismatch (Desktop upgraded):**
+1. Copy new bundled CLI to `~/.toolhive/bin/thv.new`
+2. Verify checksum
+3. Atomic rename: `thv.new` → `thv`
+4. Update marker file
+
+#### Scenario C: Existing External CLI (Homebrew/Manual)
```
peppescg: I don't think it could be safe to downgrade toolhive version 🤔 , my 2c I probably discourage at all this, so the case where there is already a CLI v0.6.0 and the UI on v0.5.1.

rfcs/THV-0020-desktop-cli-version-alignment.md
```diff
@@ -0,0 +1,854 @@
...
+- Desktop CLI still shadows Homebrew in PATH (no change)
+- Settings panel shows: "Other CLI detected: v0.6.0 at /opt/homebrew/bin/thv (shadowed)"
+
+**On user running `brew upgrade toolhive`:**
+- Homebrew upgrades its copy
+- Desktop CLI still shadows it
+- User may be confused why `thv --version` shows old version
+
+**Mitigation:** Settings panel explains PATH shadowing, links to documentation.
+
```
peppescg: IMO in this scenario it would be helpful to show something in the cli as well

rfcs/THV-0020-desktop-cli-version-alignment.md
```diff
@@ -0,0 +1,854 @@
+# CLI and Desktop App Version Alignment
+
+- **Status**: Draft
+- **Author(s)**: @samuv
+- **Created**: 2025-01-12
```
dmjb: Nitpick: Wrong year :)

rfcs/THV-0020-desktop-cli-version-alignment.md
```diff
@@ -0,0 +1,854 @@
...
+    H -->|User: Install Desktop CLI| F
+    H -->|User: Skip| K[Warn and Continue]
+    
+    F --> L[Configure PATH]
+    J --> L
+    L --> I
+    K --> I
+```
+
+**Key insight:** Package managers (Homebrew, Winget) distribute the Desktop app, but Desktop owns CLI installation regardless of how Desktop was installed. Package managers cannot guarantee exact versions, so Desktop must handle this itself.
```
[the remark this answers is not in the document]
kantord: i guesds the biggest problem is that you can use the ui from a tarball. so you can have as many parallel "installations" as you want

rfcs/THV-0020-desktop-cli-version-alignment.md
```diff
@@ -0,0 +1,854 @@
...
+| macOS | `~/.toolhive/bin/thv`, `/opt/homebrew/bin/thv`, `/usr/local/bin/thv` |
+| Linux | `~/.toolhive/bin/thv`, `~/.local/bin/thv`, `/usr/local/bin/thv` |
+| Windows | `%LOCALAPPDATA%\ToolHive\bin\thv.exe`, `%ProgramFiles%\toolhive\thv.exe` |
+
+#### Detection Information
+
+For each found binary:
+- Path to binary
+- Version (`thv --version`)
+- Checksum (SHA256)
```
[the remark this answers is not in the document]
samuv: The checksum ensures it's the official binary we bundled. And my question would be, why not?

It's cheap to compute (one SHA256 on startup), and gives us overall more reliability and the possibility to recover a corrupted file

Happy to mark it as optional/future if we want to keep it simpler, but felt like low-hanging fruit to me.

rfcs/THV-0020-desktop-cli-version-alignment.md
samuv: Depending on how we handle [this scenario](https://github.com/stacklok/toolhive-rfcs/pull/20#discussion_r2687455727), it would influence this the other way around too, when you already have Desktop installed and then install CLI via Homebrew afterward.
If we go with the "hard block" approach, we'd need the CLI to detect Desktop's presence and warn/block. That would mean CLI needs to be aware of Desktop install

dmjb (approved):

PR #23, opened 2026-01-14
RFC: CRD v1beta1 Optimization and Configuration Extraction

JAORMX:
## Summary

This RFC proposes architectural improvements to ToolHive's Kubernetes CRDs for the v1beta1 release.

### Key Changes

**New Configuration CRDs:**
- `MCPOIDCConfig` - Shared OIDC authentication config
- `MCPTelemetryConfig` - Shared OpenTelemetry/Prometheus config
- `MCPAuthzConfig` - Shared Cedar authorization policies
- `MCPAggregationToolConfig` - Per-workload tool filtering for VirtualMCPServer

**Structural Improvements:**
- Remove VirtualMCPServer's embedded `Config` field (CRD-first approach)
- Replace type discriminators with CEL-validated unions
- Remove deprecated fields (`port`, `targetPort`, `toolsFilter`)
- Consolidate MCPRegistry status to single phase + conditions

**Quality of Life:**
- Add printer columns for better `kubectl get` output
- Introduce common `LocalObjectReference` type

### Implementation Phases

1. **Phase 1**: Define all v1beta1 API types with CEL validation
2. **Phase 2**: Implement controllers (parallelizable)
3. **Phase 3**: Testing, migration tooling (skill + agent), release

### Related RFCs

- THV-0001 (OpenTelemetry) - MCPTelemetryConfig builds on this
- THV-0014 (vMCP K8s-Aware) - VirtualMCPServer changes align with this

### Breaking Change

This is a breaking change from v1alpha1. Migration tooling will be provided.

---

### DIFF
--- a/rfcs/THV-0023-crd-v1beta1-optimization.md
+++ b/rfcs/THV-0023-crd-v1beta1-optimization.md
@@ -5,14 +5,15 @@
 - **Created**: 2026-01-14
 - **Last Updated**: 2026-01-14
 - **Target Repository**: toolhive
-- **Related Issues**: N/A
+- **Related Issues**: 
+  - [#3125](https://github.com/stacklok/toolhive/issues/3125) - Simplify VMCP Configuration (an earlier example of unifying CRD/config types)
 - **Related RFCs**:
   - [THV-0001](THV-0001-otel-integration-proposal.md) - OpenTelemetry integration (MCPTelemetryConfig builds on this)
   - [THV-0014](THV-0014-vmcp-k8s-aware-refactor.md) - vMCP K8s-aware refactor (VirtualMCPServer changes must align)
 
 ## Summary
 
-This RFC proposes architectural improvements to ToolHive's Kubernetes CRDs for the v1beta1 release. The key changes include extracting shared configuration (OIDC, telemetry, authorization) into dedicated reusable CRDs, replacing type discriminators with CEL-validated unions, and removing deprecated fields. These changes follow the successful pattern established by `MCPExternalAuthConfig` and align with Kubernetes API best practices.
+This RFC proposes architectural improvements to ToolHive's Kubernetes CRDs for the v1beta1 release. The key changes include extracting shared configuration (OIDC, telemetry, authorization) into dedicated reusable CRDs, replacing type discriminators with CEL-validated unions, unifying CRD config types with application config types, and removing deprecated fields. These changes follow the successful pattern established by `MCPExternalAuthConfig` and align with Kubernetes API best practices.
 
 ## Problem Statement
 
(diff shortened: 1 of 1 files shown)


rfcs/THV-XXXX-crd-v1beta1-optimization.md
```diff
@@ -0,0 +1,1109 @@
...
+    requiredClaims:
+      iss: https://kubernetes.default.svc
+  # OR
+  configMapRef:
+    name: oidc-config
+    key: oidc.json
+  # OR
+  inline:
+    issuer: https://keycloak.example.com/realms/prod
+    audience: toolhive-platform
```
jhrozek: audience must not be shared between servers or else a token acquired against one server can be replayed against another server.

jhrozek: the example doesn't list scopes, but those should probably not be shared either.

rfcs/THV-XXXX-crd-v1beta1-optimization.md
```diff
@@ -0,0 +1,1109 @@
...
+    usernameClaim: sub
+    groupsClaim: groups
+status:
+  observedGeneration: 1
+  configHash: "sha256:abc123..."
+  referencingWorkloads:
+    - kind: MCPServer
+      name: server-a
+    - kind: MCPServer
+      name: server-b
```
jhrozek: is there any worry if this object is shared against many servers that the status will grow large? I guess we won't overflow 1MB though

rfcs/THV-XXXX-crd-v1beta1-optimization.md
```diff
@@ -0,0 +1,1109 @@
...
+        - list_repos
+        - create_issue
+      overrides:
+        search_code:
+          name: github_search  # Rename for clarity
+          description: "Search GitHub repositories for code"
+    - workload: slack-mcp
+      filter:
+        - send_message
+        - list_channels
```
jhrozek: I wonder how much it even makes sense to split out the authz CRD if we're going to include per-server items anyway?

rfcs/THV-XXXX-crd-v1beta1-optimization.md
```diff
@@ -0,0 +1,1109 @@
...
+
+#### New CRD 3: MCPAuthzConfig
+
+Shared Cedar authorization policies.
+
+```yaml
+apiVersion: toolhive.stacklok.dev/v1beta1
+kind: MCPAuthzConfig
+metadata:
+  name: standard-authz
```
jhrozek: you don't include a namespace here which might be by design..maybe we should spell it out in the doc? (I think it's the only acceptable variant otherwise you can potentially give access to modifying authorization for unintended servers)

In general I wonder if authorization should still be expressed per-server as well because the cedar policies might contain tool call names which are inherently per-server.

jhrozek (commented):
Added a bunch of comments related to the shared CRDs. I'm not sure if this proposal is not pulling in this opposite direction as @jerm-dro has been working in the vMCP config space.

rfcs/THV-XXXX-crd-v1beta1-optimization.md
JAORMX: Good catch on the security issue. I've restructured MCPOIDCConfig to only include shareable parts (issuer, jwksUri, claim mappings, client credentials). `audience` and `scopes` are now per-server fields in the reference:

```yaml
oidcConfig:
  ref:
    name: production-oidc
    namespace: mcp-platform
  audience: github-mcp-server    # per-server (required)
  scopes: [openid, github:repo]  # per-server (optional)
```

This way each server gets a unique audience, preventing token replay.

PR #26, opened 2026-01-21
Add RFC THV-0020 for header injection in remote MCP proxies

jhrozek:
This RFC proposes server-side configuration for injecting HTTP headers into requests forwarded to remote MCP servers. Currently, if specific headers need to be sent to remote servers, clients must configure them individually, which is brittle and doesn't scale across different MCP client implementations.

The solution uses a middleware that injects configured header name-value pairs into every request, following the existing token_injection pattern. This provides a single, centralized configuration point for operators.

Affected components:
- thv proxy command (--remote-forward-headers flag)
- thv run command for remote URLs (--remote-forward-headers flag)
- MCPRemoteProxy Kubernetes CRD (headerForward.addHeaders field)

Related: stacklok/toolhive#3316

### DIFF
--- a/rfcs/THV-0026-header-passthrough.md
+++ b/rfcs/THV-0026-header-passthrough.md
@@ -1,9 +1,9 @@
-# THV-0020: Configurable Header Passthrough for Remote MCP Servers
+# THV-0026: Configurable Header Passthrough for Remote MCP Servers
 
 - **Status**: Draft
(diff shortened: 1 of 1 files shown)


jhrozek:
@JAORMX I added a new section that focuses on adding headers with potentially secret value, wdyt?

PR #38, opened 2026-02-04
RFC: Session-Scoped Architecture for vMCP

yrobla:
RFC: Session-Scoped Architecture for vMCP                                                                                       
                                          
This RFC refactors vMCP session management to fix per-request client creation and scattered session concerns.                            
                                                                                                                                           
Problem                                                                                                                                  
                                                                                                                                           
  - Per-request overhead: Backend clients created/closed on every tool call, causing connection overhead and state loss for stateful
  backends (Playwright, databases)                                                                                                         
  - Scattered architecture: Session creation tangled across middleware, adapters, and SDK hooks with no clear ownership
  - Hard to extend: Current design makes it difficult to add features (optimizer-in-vmcp) or write integration tests without spinning up
  the full server

Solution

Introduce two core interfaces:

  1. Session - Domain object that owns backend clients, encapsulates routing logic, manages lifecycle
  2. SessionFactory - Creates fully-formed sessions (capability discovery + client initialization)
  3. SessionManager - Bridges domain logic to SDK protocol concerns

Key improvement: Clients created once during session init, reused for all requests, closed on expiration.

### DIFF
--- a/rfcs/THV-0038-session-scoped-client-lifecycle.md
+++ b/rfcs/THV-0038-session-scoped-client-lifecycle.md
@@ -3,7 +3,7 @@
 - **Status**: Draft
 - **Author(s)**: @yrobla, @jerm-dro
 - **Created**: 2026-02-04
-- **Last Updated**: 2026-02-12
+- **Last Updated**: 2026-02-16
 - **Target Repository**: toolhive
 - **Related Issues**: [toolhive#3062](https://github.com/stacklok/toolhive/issues/3062)
 
@@ -11,11 +11,30 @@
 
 Refactor vMCP session architecture around three core components: `Session` (domain object with behavior), `SessionFactory` (creates fully-formed sessions), and `SessionManager` (bridges domain logic to MCP SDK protocol). This separates session creation (domain logic) from SDK integration (protocol concerns) and integrates with the existing pluggable session storage architecture.
 
+**Multi-client session model**: One vMCP instance serves multiple clients (e.g., Claude Desktop, VSCode, Cursor) via a single HTTP endpoint. Each client that sends an `InitializeRequest` gets a unique session ID (per MCP Streamable HTTP protocol). Sessions are created dynamically when clients connect, not pre-created at startup. Each session owns its own backend clients for complete isolation.
+
 Restructuring the session management enables:
 
-- **Client reuse throughout the session**: Clients are created once during initialization, reused for all requests, and closed during cleanup, enabling stateful workflows (e.g., Playwright browser sessions, database transactions)
+- **Client reuse throughout the session**: Backend clients are created once during session initialization, reused for all requests within that session, and closed during cleanup, enabling stateful workflows (e.g., Playwright browser sessions, database transactions)
 - **Integration testing without full server**: Test session capabilities and routing logic in isolation without spinning up the complete vMCP server
 - **Simplified feature extensions**: Add capabilities like optimizer-in-vmcp and capability refresh through interface decoration, since protocol concerns are decoupled from how capabilities are calculated
+
+## Terminology
+
+This RFC uses the following terms consistently:
+
+| Term | Definition | Example |
+|------|------------|---------|
+| **Backend** | A deployed MCP server instance that vMCP aggregates | Filesystem MCP server, Playwright MCP server |
+| **Backend workload** | Same as "backend" - the deployment unit | Used interchangeably with "backend" |
+| **MCP server** | Generic term for any MCP-compliant server | Can refer to vMCP itself or a backend |
+| **vMCP session** | Client-to-vMCP session identified by vMCP's session ID | Session created when Claude Desktop initializes with vMCP |
+| **Backend session** | vMCP-to-backend session identified by backend's session ID | Session between vMCP's client and a Playwright backend |
+| **Backend client** | MCP client instance connecting vMCP to one backend | Instance of `mcp-go/client.Client` |
+| **Session-scoped** | Resources that live for the duration of a vMCP session | Backend clients created at init, closed at session expiration |
+| **Session ID** | Unique identifier for a vMCP session (UUID) | Returned in `Mcp-Session-Id` header |
+
+**Key distinction**: A single **vMCP session** owns multiple **backend clients**, each potentially maintaining its own **backend session** with the respective backend MCP server.
 
 ## Problem Statement
 
@@ -128,7 +147,7 @@ sequenceDiagram
     participant Hook as OnRegisterSession
     participant VMCPSess as VMCPSession
 
-    Client->>DiscoveryMW: POST /mcp/initialize
+    Client->>DiscoveryMW: POST /mcp<br/>(method: initialize)
     Note over DiscoveryMW: Capabilities discovered<br/>BEFORE session exists
     DiscoveryMW->>DiscoveryMW: Discover capabilities
     DiscoveryMW->>DiscoveryMW: Stuff into request context
@@ -162,7 +181,51 @@ This RFC proposes restructuring around three core components:
 2. **SessionFactory**: Creates fully-formed sessions with all dependencies (capability discovery, client initialization, resource allocation)
 3. **SessionManager**: Bridges domain logic (Session/SessionFactory) to MCP SDK protocol, using the existing `transportsession.Manager` for storage
 
-**Proposed session creation flow** (two-phase pattern due to SDK constraints):
+#### Multi-Client Session Model
+
+**One vMCP instance serves multiple clients**, each with their own isolated session:
+
+```
+┌─────────────┐
+│ Client A    │──┐
+│ (Claude)    │  │
+└─────────────┘  │     ┌────────────────────────────┐
+                 │     │  vMCP Server               │
+┌─────────────┐  ├────→│  Endpoint: /mcp            │
+│ Client B    │──┤     │  (Streamable HTTP)         │
+│ (VSCode)    │  │     └────────────────────────────┘
+└─────────────┘  │              │
+                 │              ├─ Session aaa-111 (Client A's session)
+┌─────────────┐  │              ├─ Session bbb-222 (Client B's session)
+│ Client C    │──┘              └─ Session ccc-333 (Client C's session)
+│ (Cursor)    │
+└─────────────┘
+```
+
+**How sessions are created per the MCP Streamable HTTP protocol:**
+
+- vMCP exposes **one HTTP endpoint** (e.g., `POST /mcp`) per [MCP Streamable HTTP spec](https://spec.modelcontextprotocol.io/specification/2025-06-18/basic/transports/)
+- **All JSON-RPC methods** (`initialize`, `tools/call`, `resources/read`, etc.) go to the **same endpoint**
+  - Method specified in JSON-RPC body, not URL path
+  - Example: `POST /mcp` with body `{"jsonrpc": "2.0", "method": "initialize", ...}`
+- **Each client** that sends an `InitializeRequest` triggers creation of a **new unique session**
+- The server returns a unique session ID via the `Mcp-Session-Id` header
+- Clients include their session ID in the `Mcp-Session-Id` header on all subsequent requests
+- **Sessions are created dynamically when clients connect**, not pre-created at vMCP startup
+- Each session owns its own backend clients (no sharing between sessions)
+
+**vMCP session vs backend sessions** (see [Terminology](#terminology) for definitions):
+
+Clients interact with a **single vMCP session ID** (e.g., `aaa-111`), but vMCP internally manages **multiple backend session IDs**—one per backend MCP server. vMCP maintains an explicit `[vMCP_session_id] -> {backend_id: backend_session_id}` mapping for observability (logging, metrics, health checks, debugging).
+
+**Key behaviors**:
+- Client objects include backend session IDs in request headers (MCP protocol requirement)
+- vMCP tracks backend session IDs separately (production observability requirement)
+- Each vMCP session owns isolated backend clients and session mappings
+
+For detailed lifecycle management (initialization, runtime, re-initialization, cleanup), see [Backend Session ID Lifecycle Management](#3-session-implementation).
+
+**Proposed session creation flow (per client, two-phase pattern due to SDK constraints):**
 
 ```mermaid
 sequenceDiagram
@@ -173,19 +236,22 @@ sequenceDiagram
     participant Aggregator
     participant Session
 
-    Client->>SDK: POST /mcp/initialize
+    Note over Client,Session: Session Creation (happens for EACH client that connects)
+
+    Client->>SDK: POST /mcp<br/>(method: initialize)
     Note over SDK: SDK handles initialize,<br/>needs session ID
     SDK->>SessionMgr: Generate() // No context!
-    Note over SessionMgr: Phase 1: Create empty<br/>session placeholder
+    Note over SessionMgr: Phase 1: Create NEW unique<br/>session ID and placeholder
     SessionMgr->>SessionMgr: Store empty session by ID
-    SessionMgr-->>SDK: sessionID
+    SessionMgr-->>SDK: sessionID (unique per client)
     SDK->>SessionMgr: OnRegisterSession(sessionID, ctx)
     Note over SessionMgr: Phase 2: Now we have context!<br/>Populate the session
     SessionMgr->>Factory: MakeSession(ctx, identity, backends)
     Factory->>Factory: Create clients for each backend
+    Note over Factory: Each client sends Initialize to<br/>its backend, receives backend's<br/>session ID (stored by client)<br/>Factory captures session IDs<br/>via client.SessionID()
     Factory->>Aggregator: AggregateCapabilities(ctx, clients)
     Aggregator-->>Factory: Capabilities (tools, resources, routing)
-    Factory->>Session: new Session(capabilities, clients)
+    Factory->>Session: new Session(capabilities, clients, backendSessionIDs)
     Session-->>Factory: session
     Factory-->>SessionMgr: session
     SessionMgr->>SessionMgr: Replace placeholder with real session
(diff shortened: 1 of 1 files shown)


rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,536 @@
...
+- **Created**: 2026-02-04
+- **Last Updated**: 2026-02-04
+- **Target Repository**: `toolhive`
+- **Related Issues**: [toolhive#3062](https://github.com/stacklok/toolhive/issues/3062)
+
+## Summary
+
+Move MCP backend client lifecycle from per-request creation/destruction to session-scoped management. Clients will be created once during session initialization, reused throughout the session lifetime, and closed during session cleanup. This simplifies the architecture and ensures consistent backend state preservation.
+
+## Problem Statement
```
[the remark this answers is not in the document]
jerm-dro: Stating the client problem another way: "it's hard & unintuitive to create objects in vMCP with a session lifetime." 

These tangled concerns produce other unfortunate effects. The first that comes to mind is missing integration tests below the server level. Because we can't create sessions without spinning up the whole vMCP server, we are stuck testing very small units (e.g. the `discovery` module) or the whole server (e.g. gingko/integration tests) Related issue: https://github.com/stacklok/toolhive/issues/2852.

I think the solution should look something like the following (generated with claude's assistance):

# The Solution: Decoupling Session Creation from SDK Wiring

## Background: How Session Routing Works Today

The current design routes requests via **context**, not session object lookup.

1. **Middleware stuffs capabilities into context** before the SDK sees the request: [discovery/middleware.go:109-110](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/discovery/middleware.go#L109-L110)

2. **Router is stateless** - it extracts the routing table from context on every request: [router/default_router.go:54](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/router/default_router.go#L54)
   ```go
   capabilities, ok := discovery.DiscoveredCapabilitiesFromContext(ctx)
   target := capabilities.RoutingTable.Tools[toolName]
   ```

3. **Handler factory uses router** to find backend target, then calls a shared backend client: [adapter/handler_factory.go:103-125](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/server/adapter/handler_factory.go#L103-L125)
   ```go
   target, err := f.router.RouteTool(ctx, toolName)
   result, err := f.backendClient.CallTool(ctx, target, toolName, args, meta)
   ```

The flow is: `request → middleware (stuff ctx) → handler → router (read ctx) → backend client`

There's no `sessions.Load(sessionID)` - routing data flows through context. The backend client is shared across all sessions and creates new MCP connections per request.

---

## Core Interfaces

Both the session factory and session itself should be interfaces for testability:

```go
// SessionFactory creates fully-formed sessions from configuration and runtime inputs.
type SessionFactory interface {
    // MakeSession constructs a session with all its dependencies.
    // This is where capability discovery, client creation, and resource allocation happen.
    MakeSession(
        ctx context.Context,
        identity *auth.Identity,
        backends []Backend,
    ) (Session, error)
}

// Session represents an active MCP session with all its capabilities and resources.
// This is a pure domain object - no protocol concerns.
type Session interface {
    ID() string

    // Capabilities - returns discovered tools/resources for this session
    Tools() []Tool
    Resources() []Resource

    // MCP operations - routing logic is encapsulated here
    CallTool(ctx context.Context, name string, arguments map[string]any) (*ToolResult, error)
    ReadResource(ctx context.Context, uri string) (*ResourceResult, error)
    GetPrompt(ctx context.Context, name string, arguments map[string]any) (*PromptResult, error)

    // Lifecycle
    Close() error
}
```

---

## Session Factory Implementation

```go
type defaultSessionFactory struct {
    aggregator    Aggregator
    clientFactory ClientFactory
    config        *Config
}

func (f *defaultSessionFactory) MakeSession(
    ctx context.Context,
    identity *auth.Identity,
    backends []Backend,
) (Session, error) {
    // 1. Discover capabilities (queries all backends, resolves conflicts)
    capabilities, err := f.aggregator.AggregateCapabilities(ctx, backends)
    if err != nil {
        return nil, fmt.Errorf("capability discovery failed: %w", err)
    }

    // 2. Create MCP clients for all backends in routing table
    clients := make(map[string]*Client)
    for _, target := range capabilities.RoutingTable.UniqueBackends() {
        c, err := f.clientFactory.CreateAndInitialize(ctx, target)
        if err != nil {
            closeAll(clients)
            return nil, fmt.Errorf("client creation failed for %s: %w", target.WorkloadID, err)
        }
        clients[target.WorkloadID] = c
    }

    // 3. Return fully-formed session
    return &defaultSession{
        id:           uuid.New().String(),
        identity:     identity,
        routingTable: capabilities.RoutingTable,
        tools:        capabilities.Tools,
        resources:    capabilities.Resources,
        clients:      clients,
    }, nil
}
```

---

## Session Implementation

The session encapsulates routing logic and holds pre-created clients:

```go
type defaultSession struct {
    id           string
    identity     *auth.Identity
    routingTable *RoutingTable
    tools        []Tool
    resources    []Resource
    clients      map[string]*Client // backendID -> client
}

func (s *defaultSession) ID() string            { return s.id }
func (s *defaultSession) Tools() []Tool         { return s.tools }
func (s *defaultSession) Resources() []Resource { return s.resources }

func (s *defaultSession) CallTool(ctx context.Context, name string, args map[string]any) (*ToolResult, error) {
    target, ok := s.routingTable.Tools[name]
    if !ok {
        return nil, fmt.Errorf("tool not found: %s", name)
    }

    client := s.clients[target.WorkloadID]
    return client.CallTool(ctx, target.OriginalName, args)
}

func (s *defaultSession) Close() error {
    var errs []error
    for id, c := range s.clients {
        if err := c.Close(); err != nil {
            errs = append(errs, fmt.Errorf("closing client %s: %w", id, err))
        }
    }
    s.clients = nil
    return errors.Join(errs...)
}
```

---

## Wiring into the MCP SDK

The mark3labs/mcp-go SDK defines [SessionIdManager](https://github.com/mark3labs/mcp-go/blob/main/server/streamable_http.go):

```go
type SessionIdManager interface {
    Generate() string
    Validate(sessionID string) (isTerminated bool, err error)
    Terminate(sessionID string) (isNotAllowed bool, err error)
}
```

A single `SessionManager` implements both SDK session lifecycle and provides adapted tools/resources:

```go
// SessionManager implements SDK session lifecycle and provides adapted capabilities.
type SessionManager struct {
    factory  SessionFactory
    sessions sync.Map // map[string]Session
}

// --- SessionIdManager interface (SDK lifecycle) ---

func (m *SessionManager) Generate(ctx context.Context) string {
    identity, _ := auth.IdentityFromContext(ctx)
    backends := getBackendsFromContext(ctx)

    session, err := m.factory.MakeSession(ctx, identity, backends)
    if err != nil {
        logger.Errorf("session creation failed: %v", err)
        return ""
    }

    m.sessions.Store(session.ID(), session)
    return session.ID()
}

func (m *SessionManager) Validate(sessionID string) (isTerminated bool, err error) {
    _, ok := m.sessions.Load(sessionID)
    if !ok {
        return false, fmt.Errorf("session not found")
    }
    return false, nil
}

func (m *SessionManager) Terminate(sessionID string) (isNotAllowed bool, err error) {
    val, ok := m.sessions.LoadAndDelete(sessionID)
    if !ok {
        return false, nil
    }

    session := val.(Session)
    if err := session.Close(); err != nil {
        logger.Warnf("error closing session %s: %v", sessionID, err)
    }
    return false, nil
}

// --- Adapted capabilities for SDK registration ---

func (m *SessionManager) GetAdaptedTools(sessionID string) []mcp.Tool {
    val, ok := m.sessions.Load(sessionID)
    if !ok {
        return nil
    }
    session := val.(Session)

    var sdkTools []mcp.Tool
    for _, tool := range session.Tools() {
        sdkTools = append(sdkTools, mcp.NewTool(
            tool.Name,
            m.createToolHandler(sessionID, tool.Name),
            mcp.WithDescription(tool.Description),
            mcp.WithToolInputSchema(tool.InputSchema),
        ))
    }
    return sdkTools
}

func (m *SessionManager) GetAdaptedResources(sessionID string) []mcp.Resource {
    val, ok := m.sessions.Load(sessionID)
    if !ok {
        return nil
    }
    session := val.(Session)

    var sdkResources []mcp.Resource
    for _, res := range session.Resources() {
        sdkResources = append(sdkResources, mcp.NewResource(
            res.URI,
            m.createResourceHandler(sessionID, res.URI),
            mcp.WithResourceDescription(res.Description),
            mcp.WithMIMEType(res.MimeType),
        ))
    }
    return sdkResources
}

// --- Internal handler creation ---

func (m *SessionManager) createToolHandler(sessionID, toolName string) mcp.ToolHandler {
    return func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
        val, ok := m.sessions.Load(sessionID)
        if !ok {
            return nil, fmt.Errorf("session not found: %s", sessionID)
        }
        session := val.(Session)

        args, _ := req.Params.Arguments.(map[string]any)
        result, err := session.CallTool(ctx, toolName, args)
        if err != nil {
            return mcp.NewToolResultError(err.Error()), nil
        }

        return toSDKResult(result), nil
    }
}

func (m *SessionManager) createResourceHandler(sessionID, uri string) mcp.ResourceHandler {
    return func(ctx context.Context, req mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) {
        val, ok := m.sessions.Load(sessionID)
        if !ok {
            return nil, fmt.Errorf("session not found: %s", sessionID)
        }
        session := val.(Session)

        result, err := session.ReadResource(ctx, uri)
        if err != nil {
            return nil, err
        }

        return toSDKResourceContents(result), nil
    }
}
```

Registration with the SDK:

```go
// 1. Session lifecycle management
streamableServer := server.NewStreamableHTTPServer(
    mcpServer,
    server.WithSessionIdManager(sessionManager),
)

// 2. In OnRegisterSession hook, register adapted tools/resources
hooks.AddOnRegisterSession(func(ctx context.Context, sdkSession server.ClientSession) {
    // The SDK calls sessionManager.Generate() first, which creates the session and
    // returns its ID. The SDK then fires this hook, passing that same ID via
    // sdkSession.SessionID(). This allows us to look up the session we created
    // in Generate() and register its tools/resources with the SDK.
    sessionID := sdkSession.SessionID()

    mcpServer.AddSessionTools(sessionID, sessionManager.GetAdaptedTools(sessionID)...)
    mcpServer.AddSessionResources(sessionID, sessionManager.GetAdaptedResources(sessionID)...)
})
```

---

## Request Routing via Session Lookup

**Current flow**: `request → middleware (stuff ctx) → handler → router (read ctx) → shared backend client`

**Proposed flow**: `request → handler (captures sessionID) → sessions.Load(id) → session.CallTool() → session's client`

The `SessionManager` owns the sessions map and provides:
- `SessionIdManager` methods for SDK lifecycle (`Generate`, `Validate`, `Terminate`)
- `GetAdaptedTools`/`GetAdaptedResources` for SDK registration with handlers that route to sessions

---

## Client Lifecycle Becomes Trivial

| Event | Location | What Happens |
|-------|----------|--------------|
| Session created | `SessionFactory.MakeSession()` | All clients created and initialized |
| Tool call | `Session.CallTool()` | Uses pre-created client from internal map |
| Session ends | `Session.Close()` via `Terminate()` | All clients closed |

No lazy initialization, no pools, no decorators, no cleanup gaps.

---

## Testability

### Testing Session Logic (No Server Required)

```go
func TestSession_CallTool(t *testing.T) {
    mockClient := &MockClient{
        CallToolFunc: func(ctx context.Context, name string, args map[string]any) (*ToolResult, error) {
            return &ToolResult{Content: "success"}, nil
        },
    }

    session := &defaultSession{
        id: "test-session",
        routingTable: &RoutingTable{
            Tools: map[string]*BackendTarget{
                "fetch": {WorkloadID: "backend-1", OriginalName: "fetch"},
            },
        },
        clients: map[string]*Client{"backend-1": mockClient},
    }

    result, err := session.CallTool(ctx, "fetch", map[string]any{"url": "https://example.com"})

    assert.NoError(t, err)
    assert.Equal(t, "success", result.Content)
}
```

### Testing Session Factory

```go
func TestSessionFactory_MakeSession(t *testing.T) {
    mockAggregator := &MockAggregator{...}
    mockClientFactory := &MockClientFactory{...}

    factory := &defaultSessionFactory{
        aggregator:    mockAggregator,
        clientFactory: mockClientFactory,
    }

    identity := &auth.Identity{Subject: "user-123"}
    backends := []Backend{{ID: "backend-1"}}

    session, err := factory.MakeSession(ctx, identity, backends)

    assert.NoError(t, err)
    assert.NotEmpty(t, session.ID())
}
```

### Testing SessionManager in Isolation

```go
func TestSessionManager_Terminate_ClosesSession(t *testing.T) {
    closeCalled := false
    mockSession := &MockSession{
        IDFunc:    func() string { return "session-123" },
        CloseFunc: func() error { closeCalled = true; return nil },
    }

    manager := &SessionManager{}
    manager.sessions.Store("session-123", mockSession)

    isNotAllowed, err := manager.Terminate("session-123")

    assert.NoError(t, err)
    assert.False(t, isNotAllowed)
    assert.True(t, closeCalled)

    _, exists := manager.sessions.Load("session-123")
    assert.False(t, exists)
}
```

jerm-dro (commented):
I left two bigger comments elaborating on my thoughts. To summarize, I think there are larger problems  with how the session concept is handled within vMCP. Namely, we have session concerns scattered throughout the codebase and all of these concerns are tightly coupled. As a solution, I think we should create an interface that encapsulates these concerns:

```go
// Session represents an active MCP session with all its capabilities and resources.
// This is a pure domain object - no protocol concerns and can be run without being spun up in a server.
type Session interface {
    ID() string

    // Capabilities - returns discovered tools/resources for this session
    // Perhaps these could be combined into one "GetCapabilities" method.
    Tools() []Tool
    Resources() []Resource

    // MCP operations - routing logic is encapsulated here
    CallTool(ctx context.Context, name string, arguments map[string]any) (*ToolResult, error)
    ReadResource(ctx context.Context, uri string) (*ResourceResult, error)
    GetPrompt(ctx context.Context, name string, arguments map[string]any) (*PromptResult, error)

    // Lifecycle
    Close() error
}
```

That would make solving this client-lifetime problem trivial.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,924 @@
...
+
+    // Lifecycle
+    Close() error
+}
+```
+
+**Key behaviors:**
+- **Owns backend clients**: Session stores pre-initialized MCP clients in an internal map
+- **Encapsulates routing**: `CallTool()` looks up tool in routing table, routes to correct backend client
+- **Manages lifecycle**: `Close()` iterates through all clients and closes them
```
jerm-dro: Nit: I wouldn't say "manages lifecycle" but instead say "Manageable Lifetime - Close() cleans up clients and any other resources. The SessionManager / caller is decoupled from what excatly happens on close()."

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,924 @@
...
+   - Clients used for discovery are already working and initialized
+
+3. **Return fully-formed session**: Construct session with all dependencies
+   - Session ID (UUID)
+   - User identity
+   - Routing table (maps tool names to backend workload IDs)
+   - Tools, resources, prompts (aggregated from all backends)
+   - Pre-initialized clients map (keyed by workload ID)
+
+**Key Design Decision**: The aggregator returns both capabilities and clients together. This enables the factory to pass initialized clients directly to the session without additional connection overhead.
```
jerm-dro: Nit: It'd be cleaner to 1) create the clients 2) pass them to discovery 3) pass them to the session.

There's no reason discovery has to create the clients.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,924 @@
...
+**Current workaround**: The existing `sessionIDAdapter.Generate()` creates an **empty session** (just a UUID), then `OnRegisterSession` hook populates it later. This is the "tangled flow" we're trying to fix.
+
+**The Challenge**: Ideally, discovery middleware (which has request context) would create the fully-formed session, and `Generate()` would just return the ID. However, Go lacks goroutine-local storage, so `Generate()` can't retrieve a pre-created session ID without unsafe patterns.
+
+**Implementation Options**:
+
+The SDK constraint requires choosing between these approaches:
+
+**Option A: Hybrid approach (minimal changes to current flow)** ⭐ **Recommended for Phase 1**
+- Discovery middleware discovers capabilities and stores in context (current behavior)
```
jerm-dro: Can't we pull this responsibility into the `SessionFactory`?

jerm-dro (commented):
This is coming together well! Most comments are minor or to clarify my understanding.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -358,30 +400,19 @@ type SessionIdManager interface {
...
 
-**Option A: Hybrid approach (minimal changes to current flow)** ⭐ **Recommended for Phase 1**
-- Discovery middleware discovers capabilities and stores in context (current behavior)
-- `Generate()` creates session ID, stores empty session
-- `OnRegisterSession` hook retrieves capabilities from context and calls `SessionFactory.MakeSession()`
-- Replace empty session with fully-formed session in SessionManager
-- **Pros**: Works with SDK as-is, minimal risk, no unsafe patterns
-- **Cons**: Still uses context-passing and hook pattern (not fully encapsulated)
+1. **Phase 1 - Session ID generation**: `Generate()` creates session ID, stores empty session placeholder
+2. **Phase 2 - Session population**: `OnRegisterSession` hook retrieves capabilities from request context and calls `SessionFactory.MakeSession()` to create the fully-formed session, then replaces the placeholder
```
jerm-dro: This "hook retrieves capabilities from request context" seems to contradict the "SessionFactory Implementation" design.  

Can you be more specific than capabilities? Based on the interface to MakeSession, I think you just mean backends and identity?

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -439,14 +472,14 @@ With the recommended Option A implementation:
...
+     - Keep session in map (tool calls will check `session.InitError()`)
    - Subsequent tool calls to this session:
-     - Check session metadata for `init_error`
-     - Return clear error: "Session initialization failed: [details]"
+     - Check `session.InitError()` before processing request
+     - If non-nil, return clear error: "Session initialization failed: [details from InitError()]"
      - Client should delete session and re-initialize
 
-**Rationale**: Option A uses a two-phase creation (empty session + populate via hook). The SDK's `Generate()` must return a session ID, so failures during `MakeSession()` are handled by storing error state in the session object itself.
+**Rationale**: The two-phase creation pattern (empty session + populate via hook) is necessary because the SDK's `Generate()` must return a session ID. Failures during `MakeSession()` are handled by storing error state in the session via `InitError()`. The Session interface includes `InitError() error` to support this.
```
jerm-dro: I see `OnRegisterSession` doesn't allow you to return an error. You should note that, as it's unclear why we have to do this dance with saving errors.

Aside, out of scope: this sdk has quite clunky interfaces and I wonder if we should build / use something better.

jerm-dro (commented):
After you address my remaining comments, can you update the PR description and share with the team for a review?

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+- No automatic client re-initialization (clients live for session lifetime)
+- Health monitoring tracks backend health (existing behavior)
+
+**Client Closed Mid-Request**:
+
+Race condition exists: session may be terminated while a request is in flight:
+- Session `Close()` closes all backend clients
+- In-flight requests receive "client closed" error from MCP library
+- Mitigation: Session storage layer extends TTL on every request (via existing `Get()` call which touches session), reducing race window
+- Future enhancement: Add reference counting to delay `Close()` until in-flight requests complete
```
ChrisJBurns: I don't think the `RWMutex` alone will prevent the closing during inflight race. If TTL cleanup fires `Close()` while a `CallTool()` is mid network round trip, the next `CallTool()` after `Close()` completes will use a closed client? `RWMutex` serializes access but doesn't convey "session is dead." After `Close()` releases the write lock, a subsequent `CallTool()` can acquire a read lock and use a closed client.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+- **Encapsulates routing**: `CallTool()` looks up tool in routing table, routes to correct backend client
+- **Manageable lifetime**: `Close()` cleans up clients and any other resources. The SessionManager/caller is decoupled from what exactly happens on close()
+- **Thread-safe**: All methods protected by internal RWMutex for concurrent access. Read operations (Tools, Resources, Prompts, CallTool, ReadResource, GetPrompt) use read locks; write operations (Close) use write locks. Methods like `Tools()`, `Resources()`, `Prompts()` return defensive copies (not references to internal slices) to prevent caller mutations. Internal maps and slices are never exposed directly. Proper encapsulation may allow removal of mutexes elsewhere (e.g., SessionManager, current VMCPSession) since state is now owned by the Session
+
+**Separation of concerns**: The Session interface focuses on domain logic (capabilities, routing, client ownership). TTL management and storage are handled by the existing session storage layer (`pkg/transport/session`) - sessions are stored in a storage backend that automatically extends TTL on access via `Get()` and expires sessions based on configured TTL.
```
ChrisJBurns: `Storage.DeleteExpired()` in` pkg/transport/session/manager.go` only deletes from storage, rather than calling `Close()`. This seems like a resource leak because expired sessions' backend connections are never closed.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+- Returns fully-formed session ready for use
+
+#### 2. SessionFactory Implementation
+
+**Implementation approach** (`pkg/vmcp/session/factory.go`):
+
+The default factory implementation follows this pattern:
+
+1. **Create MCP clients**: Initialize clients for each backend
+   - Factory creates one client per backend using existing client factory
```
ChrisJBurns: Will it do this in parallel?  For 20 backends, at ~100-500ms each, sequential initialisation means 2-10 seconds of session creation latency. I think this RFC should specify parallel client creation (even if it is a future thing) using something like `errgroup` with bounded concurrency and a per-backend timeout so one slow backend doesn't block the entire session.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+   - **Current state**: Circuit breaker and health monitoring exist at backend level (across all sessions) in `pkg/vmcp/health`
+   - **Challenge**: Clients are now session-scoped, but circuit breaker state must be shared across sessions
+   - **Proposed approach**:
+     - Circuit breaker and health check state remains at **backend/aggregator level** (inter-session, shared state)
+     - During `SessionFactory.MakeSession()`, consult health monitor to exclude unhealthy/circuit-broken backends
+     - Individual sessions don't maintain circuit breaker state - they call through to a shared health monitor service
+     - When a tool call fails, session delegates error reporting to the health monitor which updates shared circuit breaker state
+     - Existing health monitoring system (`pkg/vmcp/health`) continues tracking backend health independently of sessions
+   - **Open design questions**:
+     - Should sessions created before a backend is circuit-broken continue using that backend's client, or should they dynamically check health state on each call?
```
ChrisJBurns: Consider marking backends unhealthy in session state returning clean "backend unavailable" errors and lazy reconnection for recovered backends.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+- Mitigation: Session storage layer extends TTL on every request (via existing `Get()` call which touches session), reducing race window
+- Future enhancement: Add reference counting to delay `Close()` until in-flight requests complete
+
+**Session Not Found**:
+
+If client uses expired/invalid session ID:
+- Return clear error: "session not found" or "session expired"
+- Client should re-initialize via `/mcp/initialize` endpoint to create new session
+
+## Security Considerations
```
ChrisJBurns: How are session Ids bound to clients? If a sessionId leaks can another client use it? I can't see stuff around binding sessions to original client identity to prevent session hijacking? Is there anything we can do here?

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -0,0 +1,991 @@
...
+   - Returns session ID to SDK
+   - SDK sends `Mcp-Session-Id` header in `InitializeResult`
+
+2. **`OnRegisterSession` hook phase**: Calls `SessionFactory.MakeSession()` - failure possible here
+   - If `MakeSession()` fails completely (e.g., all backends unreachable):
+     - Log error with session ID and failure details
+     - Create a failed session that stores the initialization error internally
+     - Keep session in map (tool calls will check `session.InitError()`)
+   - Subsequent tool calls to this session:
+     - Check `session.InitError()` before processing request
```
ChrisJBurns: Maybe not a problem as much as its more dev based. But if we are proposing in this RFC that when a session setup fails, the error gets stored inside the session object, then every time someone tries to use the session the code has to remember to check `session.InitError()` first before doing anything?

This may be easy to forget? If someone 6 months from now adds a new method to the session interface and they forgot to add the `session.InitError()` check  at the top, that method will try to use a potentially broken session?

If this is seen as a problem too by the authors, a simpler fix is potentially not to advertise tools from backends that failed and because they don't exist to the Client, nobody tries to call them making the `InitError()` redundant

JAORMX (commented):
## So, about the session storage architecture...

Hey, appreciate the work on this RFC. The per-request client creation problem is very real and worth solving (I filed https://github.com/stacklok/toolhive/issues/2417 about exactly this pain point). The session-scoped client lifecycle and the `Session`-as-domain-object approach are solid improvements. That said, I have some concerns about how this interacts with the existing session storage architecture we built specifically for externalizing sessions to Redis/Valkey.

Let me dig in...

### What we already have in place

Over PRs [#1677](https://github.com/stacklok/toolhive/pull/1677), [#1989](https://github.com/stacklok/toolhive/pull/1989), [#1770](https://github.com/stacklok/toolhive/pull/1770), and [#1771](https://github.com/stacklok/toolhive/pull/1771), we built a deliberate, phased session management architecture:

- **`Storage` interface** (`pkg/transport/session/storage.go`): A 5-method pluggable abstraction (`Store`, `Load`, `Delete`, `DeleteExpired`, `Close`) designed specifically to swap between in-memory and distributed backends (Redis/Valkey)
- **`LocalStorage`**: In-memory `sync.Map` implementation behind the `Storage` interface
- **`Manager`**: Uses `Storage` + `Factory`, with `NewManagerWithStorage()` for injecting custom backends
- **`serialization.go`**: JSON ser/deser already prepared for Redis/Valkey (currently `nolint:unused`, ready for when we implement a distributed backend)

The whole point of this layering was: when we're ready for distributed sessions, we implement a `RedisStorage` (or equivalent) satisfying the `Storage` interface, and **nothing else changes**. Not the Manager, not the proxies, not the vMCP server.

With this in mind, here are my concerns:

### The parallel storage path problem

The RFC's `SessionManager` introduces its own `sync.Map` for session storage:

```go
// From RFC: SessionManager
type sessionManager struct {
    sessions sync.Map  // bypasses the Storage interface
    factory  SessionFactory
}
```

So, basically, this creates a **parallel session management path** that doesn't go through `transportsession.Manager` or the `Storage` interface. The RFC says _"TTL management and storage are handled by the existing session storage layer"_, but the proposed implementation stores sessions in its own `sync.Map`, not through the existing `Storage` interface.

This means when we implement Redis/Valkey for transport sessions, the vMCP session data managed by this new `SessionManager` **won't be externalized**. We'd need to separately refactor this `sync.Map` later, which is exactly the kind of rework the `Storage` interface was designed to prevent.

**Suggestion**: The `SessionManager` should delegate session storage to `transportsession.Manager` (which already uses the pluggable `Storage` interface) rather than maintaining its own `sync.Map`. The existing `Manager.AddSession()`, `Manager.Get()`, and `Manager.Delete()` methods already provide exactly what's needed here.

### How do the two Session interfaces compose?

The RFC proposes a new `Session` interface (behavior-oriented: `CallTool()`, `ReadResource()`, etc.) while we already have `transportsession.Session` (storage-oriented: `ID()`, `Type()`, `Touch()`, `GetData()`). These serve different purposes, but the RFC doesn't show how they compose.

Note that currently `VMCPSession` embeds `StreamableSession` which implements `transportsession.Session`. That's what makes it compatible with the `Storage` interface and `Manager`. The new behavior-oriented `Session` needs to either:

1. **Embed/implement** `transportsession.Session` so it remains compatible with the existing storage layer, or
2. **Clearly document** the two-layer model: transport session (storage) wraps/references domain session (behavior)

Without this, the new `Session` type won't be storable through the `Storage` interface, which breaks the externalization path.

### Partial initialization and serializability

The RFC allows sessions to continue with partial backends (some failed during `MakeSession()`). If we later add Redis/Valkey persistence, a session restored on a different instance wouldn't know which backends originally failed vs. which were never attempted. The session state model needs to account for this if it's going to be serializable.

### What I think should change

1. **Wire through `transportsession.Manager`** instead of a standalone `sync.Map`. The `Manager` already supports custom factories and pluggable storage.
2. **Show the composition model**: How does the new behavior-oriented `Session` relate to `transportsession.Session`? Does it embed it? Wrap it? Replace it?
3. **Add a section on session serializability**: What part of the new `Session` is metadata (storable in Redis/Valkey) vs. runtime state (client connections, goroutines that must stay in-memory)? Issue https://github.com/stacklok/toolhive/issues/2417 already sketched this dual-layer model.

The RFC solves a real problem and the domain-object approach is the right direction. I just want to make sure we don't accidentally create architectural debt that blocks the Redis/Valkey externalization path we've been building toward.

rfcs/THV-0038-session-scoped-client-lifecycle.md
```diff
@@ -162,7 +181,40 @@ This RFC proposes restructuring around three core components:
 2. **SessionFactory**: Creates fully-formed sessions with all dependencies (capability discovery, client initialization, resource allocation)
 3. **SessionManager**: Bridges domain logic (Session/SessionFactory) to MCP SDK protocol, using the existing `transportsession.Manager` for storage
 
-**Proposed session creation flow** (two-phase pattern due to SDK constraints):
+#### Multi-Client Session Model
+
+**One vMCP instance serves multiple clients**, each with their own isolated session:
```
amirejaz: How does each client obtain its session_id?

- Do clients need to initialize with vMCP, with vMCP returning the session_id?
- If we have multiple SSE MCP servers, each with its own session_id, how do we handle that? 

One approach: treat the vMCP session_id as the primary session, and have vMCP create/obtain a separate session_id per downstream SSE MCP server, then maintain a mapping like:

[vMCP_session_id][sse_session_id]

So the client only deals with the vMCP session_id, and vMCP hides the complexity of multiple downstream session IDs.

jerm-dro (approved):

JAORMX (approved):

PR #59, opened 2026-03-24
RFC: First-class vMCP support in the local CLI experience

JAORMX:
## Summary

- Proposes integrating vMCP into the main `thv` CLI as a `thv vmcp` subcommand (serve, validate, init)
- Brings the optimizer to local with three tiers: FTS5-only (`--optimizer`), managed TEI container (`--optimizer-embedding`), and full config control
- Formalizes the library embedding pattern used by brood-box as an officially supported integration path
- Preserves the standalone `vmcp` binary for K8s deployments and backwards compatibility

## Key Design Decisions

- **Zero-config quickstart**: `thv vmcp serve --group default` works without a config file
- **Shared logic in `pkg/vmcp/cli/`**: Both `thv vmcp` and standalone `vmcp` delegate to the same package
- **Managed TEI lifecycle**: `--optimizer-embedding` auto-manages a TEI container with named volume for model caching, health polling, and graceful fallback to FTS5-only
- **Library stability table**: Packages rated Stable/Beta/Internal to guide downstream consumers

## Related RFCs

- THV-0008 (Virtual MCP Server)
- THV-0014 (K8s-Aware vMCP)
- THV-0022 (Optimizer Migration)
- THV-0034 (Long-Running Local Server)

### DIFF
--- a/rfcs/THV-0059-vmcp-local-experience.md
+++ b/rfcs/THV-0059-vmcp-local-experience.md
@@ -1,4 +1,4 @@
-# RFC-XXXX: First-Class vMCP Support in the Local CLI Experience
+# RFC-0059: First-Class vMCP Support in the Local CLI Experience
 
 - **Status**: Draft
 - **Author(s)**: Juan Antonio Osorio (@JAORMX)
--- a/rfcs/THV-0059-vmcp-local-experience.md
+++ b/rfcs/THV-0059-vmcp-local-experience.md
@@ -3,7 +3,7 @@
 - **Status**: Draft
 - **Author(s)**: Juan Antonio Osorio (@JAORMX)
 - **Created**: 2026-03-24
-- **Last Updated**: 2026-03-24
+- **Last Updated**: 2026-04-14
 - **Target Repository**: toolhive
 - **Related Issues**: N/A
 
@@ -33,7 +33,8 @@ All ToolHive users who manage more than one MCP server locally are affected. The
 
 - Add a `thv vmcp` subcommand with `serve`, `validate`, and `init` sub-commands that integrate vMCP into the main CLI
 - Support a zero-config quickstart: `thv vmcp serve --group <name>` should work without a config file for simple aggregation
-- Bring the optimizer to the local experience with managed embedding service lifecycle (`--optimizer` flag auto-manages a TEI container)
+- Bring the optimizer to the local experience with managed embedding service lifecycle (`--optimizer-embedding` flag auto-manages a TEI container)
+- Deprecate and eventually remove the legacy standalone `mcp-optimizer` Python project (`StacklokLabs/mcp-optimizer`) once the native vMCP optimizer is stable
 - Maintain full feature parity with the standalone `vmcp` binary for users who need advanced configuration
 - Include vMCP in goreleaser so it ships with every ToolHive release (embedded in `thv` as a subcommand)
 - Document the library embedding pattern used by brood-box as an officially supported integration path
@@ -137,7 +138,7 @@ Flags:
       --optimizer                Enable the tool optimizer (FTS5 keyword search)
       --optimizer-embedding      Enable the optimizer with semantic search (auto-manages TEI container)
       --embedding-model string   HuggingFace model for semantic search (default: "BAAI/bge-small-en-v1.5")
-      --embedding-port int       Port for the embedding service container (default: 8384)
+      --embedding-image string   TEI container image (default: "ghcr.io/huggingface/text-embeddings-inference:cpu-latest")
 ```
 
 Two modes of operation:
@@ -197,16 +198,9 @@ Example output:
 name: "engineering-vmcp"
 groupRef: "engineering"
 
-# Backends discovered from group "engineering":
-# (These are auto-discovered at runtime from the group.
-#  Uncomment and modify only if you need static overrides.)
-# backends:
-#   - name: "github"
-#     url: "http://localhost:8080/mcp"
-#     transport: "streamable-http"
-#   - name: "jira"
-#     url: "http://localhost:8081/mcp"
-#     transport: "streamable-http"
+# Backends are auto-discovered at runtime from the group.
+# To modify backends, use `thv run` flags or group configuration
+# rather than overriding them here.
 
 aggregation:
   conflictResolution: prefix
@@ -311,7 +305,7 @@ type ServeConfig struct {
     Optimizer          bool   // Enable FTS5-only optimizer
     OptimizerEmbedding bool   // Enable optimizer with managed TEI container
     EmbeddingModel     string // HuggingFace model ID (default: BAAI/bge-small-en-v1.5)
-    EmbeddingPort      int    // TEI container port (default: 8384)
+    EmbeddingImage     string // TEI container image (default: ghcr.io/huggingface/text-embeddings-inference:cpu-latest)
 }
 
 // Serve starts a vMCP server with the given configuration.
@@ -419,8 +413,11 @@ thv vmcp serve --group default --optimizer-embedding
 
 # With custom model:
 thv vmcp serve --group default --optimizer-embedding \
-  --embedding-model BAAI/bge-small-en-v1.5 \
-  --embedding-port 8384
+  --embedding-model BAAI/bge-small-en-v1.5
+
+# With a GPU-accelerated image:
+thv vmcp serve --group default --optimizer-embedding \
+  --embedding-image ghcr.io/huggingface/text-embeddings-inference:turing-latest
 ```
 
 Lifecycle:
@@ -435,11 +432,11 @@ sequenceDiagram
 
     User->>THV: --optimizer-embedding
     THV->>Runtime: Start TEI container<br/>(ghcr.io/huggingface/text-embeddings-inference:cpu-latest)
-    Runtime->>TEI: Running on :8384
+    Runtime->>TEI: Running on :{random port}
     THV->>TEI: GET /health (poll until ready)
     Note over THV,TEI: TEI downloads model on first start<br/>(cached in named volume)
     TEI-->>THV: 200 OK
-    THV->>VMCP: Start with optimizer.embeddingService=http://localhost:8384
+    THV->>VMCP: Start with optimizer.embeddingService=http://localhost:{port}
     VMCP-->>User: Listening on :4483/mcp
 
     Note over User,VMCP: On shutdown (Ctrl+C / SIGTERM):
@@ -450,14 +447,14 @@ sequenceDiagram
 
 Implementation details:
 
-- **Container image**: `ghcr.io/huggingface/text-embeddings-inference:cpu-latest` (same image the K8s `EmbeddingServer` CRD uses)
-- **Container name**: `thv-embedding-<group>` (predictable, enables idempotent start/stop)
+- **Container image**: `ghcr.io/huggingface/text-embeddings-inference:cpu-latest` by default (same image the K8s `EmbeddingServer` CRD uses). Configurable via `--embedding-image` to support GPU-accelerated variants (e.g., `ghcr.io/huggingface/text-embeddings-inference:turing-latest` for CUDA) or architecture-specific images.
+- **Container name**: `thv-embedding-<model-short-hash>` (e.g., `thv-embedding-bge-sm-a1b2c3`), derived from a hash of the model name. This allows multiple vMCP instances using the same embedding model to share a single TEI container, avoiding unnecessary duplication. Different models get separate containers.
 - **Model cache**: Named volume `thv-embedding-model-cache` mounted at `/data` with `HF_HOME=/data`. This avoids re-downloading the model on every start (~130MB for `bge-small-en-v1.5`).
 - **Health check**: Poll `GET /health` with backoff until the TEI server reports ready. TEI must download and load the model on first start, which can take 30-60 seconds.
-- **Port binding**: Default `127.0.0.1:8384`. Chosen to avoid conflicting with the vMCP port (4483) or common dev ports.
-- **Lifecycle coupling**: The TEI container is started before the vMCP server and stopped after it shuts down. If the TEI container fails to start or becomes unhealthy, vMCP falls back to FTS5-only mode with a warning.
-- **Idempotent start**: If a `thv-embedding-<group>` container is already running (e.g., from a previous invocation), reuse it rather than creating a new one.
-- **Platform considerations**: The TEI CPU image is amd64. On ARM64 hosts (Apple Silicon), the container runs under emulation. A future enhancement could detect architecture and select an appropriate image variant.
+- **Port binding**: A random available port is allocated using ToolHive's existing `pkg/networking.FindAvailable()` pattern (the same mechanism `thv run` uses when no explicit port is given). The allocated port is reported in the logs. This avoids conflicts when multiple vMCP instances or other services are running.
+- **Lifecycle coupling**: The TEI container is started before the vMCP server and stopped after it shuts down. If the TEI container fails to start or become healthy when `--optimizer-embedding` was explicitly requested, `thv vmcp serve` exits with a clear error message. The `--optimizer-embedding` flag is a clear signal that the user wants semantic search — silently degrading to FTS5-only would mask environment issues (Docker not running, port conflicts, etc.) and produce confusingly poor results. Users who want keyword-only search should use `--optimizer` explicitly.
+- **Idempotent start**: If a TEI container with the matching name and ToolHive labels is already running (e.g., from a previous invocation or another vMCP instance with the same model), reuse it rather than creating a new one.
+- **Platform considerations**: The TEI CPU image is amd64-only. On ARM64 hosts (Apple Silicon), Docker Desktop handles this automatically via Rosetta 2 emulation — no `--platform` flag is needed. The overhead is roughly 5-15% for CPU-bound workloads, which is acceptable for embedding generation. The `--embedding-image` flag lets users select architecture-native images when available, making architecture management the user's responsibility rather than requiring auto-detection logic.
 
 **Tier 3: Full config control**
 
@@ -496,17 +493,15 @@ A new component in `pkg/vmcp/cli/` manages the TEI container lifecycle:
 ```go
 // EmbeddingServiceConfig holds parameters for the managed embedding container.
 type EmbeddingServiceConfig struct {
-    Image    string // default: ghcr.io/huggingface/text-embeddings-inference:cpu-latest
-    Model    string // default: BAAI/bge-small-en-v1.5
-    Port     int    // default: 8384
-    GroupRef string // used in container name: thv-embedding-<group>
+    Image string // default: ghcr.io/huggingface/text-embeddings-inference:cpu-latest
+    Model string // default: BAAI/bge-small-en-v1.5
 }
 
 // EmbeddingServiceManager manages the lifecycle of a local TEI container.
 type EmbeddingServiceManager struct { ... }
 
 // Start launches (or reuses) the TEI container and waits for readiness.
-// Returns the embedding service URL.
+// Returns the embedding service URL (with a dynamically allocated port).
 func (m *EmbeddingServiceManager) Start(ctx context.Context) (string, error) { ... }
 
 // Stop gracefully stops the TEI container.
@@ -514,6 +509,8 @@ func (m *EmbeddingServiceManager) Stop(ctx context.Context) error { ... }
 ```
 
 This uses ToolHive's existing container runtime abstraction (`pkg/container/`) to manage the TEI container, keeping it consistent with how `thv run` manages MCP server containers.
+
+**Container labels for lifecycle ownership**: The TEI container is labeled using ToolHive's existing `pkg/labels/` system — specifically the `toolhive` label (marks it as ToolHive-managed) and the `toolhive-auxiliary` label (marks it as an auxiliary workload, similar to how the inspector container is labeled). This lets `thv` query by label to determine whether an existing TEI container is ToolHive-managed (safe to reuse/stop) or user-managed (hands off). On startup, the manager checks for a running container matching the expected name and labels — if one exists and is healthy, it reuses it; if one exists without the `toolhive` label, it assumes the user is managing their own TEI and does not touch it.
 
 #### API Changes
 
@@ -682,6 +679,7 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 - **Outgoing auth**: Per-backend authentication strategies (header injection, token exchange, unauthenticated) configured via the config file. Quick mode defaults to unauthenticated.
 - **Cedar authorization**: When configured, Cedar policies gate access to individual tools and resources. The library embedding path (brood-box) demonstrates this with `observe`, `safe-tools`, and `full-access` profiles.
 - **No privilege escalation**: The `thv vmcp` subcommand runs with the same privileges as the invoking user. No root or elevated permissions required.
+- **Unix socket transport (future)**: For local single-user scenarios, a Unix socket transport (similar to `thv serve --socket`) would sidestep port-binding and auth concerns entirely, relying on filesystem permissions for access control. This is a natural follow-up but out of scope for this RFC.
 
 ### Data Security
 
@@ -718,7 +716,7 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 - Quick mode generates a conservative default config (anonymous auth, prefix conflict resolution) that is safe for local single-user scenarios.
 - Library consumers are guided toward infrastructure encapsulation and anti-corruption layers, reducing the risk of misusing internal vmcp APIs.
 - The managed TEI container binds to `127.0.0.1` only, uses the same container isolation as other ToolHive workloads, and downloads models exclusively from HuggingFace Hub (verified models).
-- If the TEI container becomes unhealthy, vMCP degrades gracefully to FTS5-only search rather than failing entirely.
+- If the TEI container fails to start when `--optimizer-embedding` was explicitly requested, `thv vmcp serve` exits with a clear error rather than silently degrading. Users who want keyword-only search should use `--optimizer` instead.
 
 ## Alternatives Considered
 
@@ -783,9 +781,9 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 ### Phase 4: Local Optimizer Support
 
 - Implement `--optimizer` flag (FTS5-only, no container management — config wiring only)
-- Implement `EmbeddingServiceManager` in `pkg/vmcp/cli/` for TEI container lifecycle
-- Implement `--optimizer-embedding` flag with container start/stop, health polling, and graceful fallback
-- Add `--embedding-model` and `--embedding-port` flags
+- Implement `EmbeddingServiceManager` in `pkg/vmcp/cli/` for TEI container lifecycle with `pkg/labels/` integration for ownership tracking
+- Implement `--optimizer-embedding` flag with container start/stop, health polling, and fail-fast on TEI failure
+- Add `--embedding-model` and `--embedding-image` flags
 - Add named volume management for model caching
 - Add E2E tests for optimizer tiers (FTS5-only, managed TEI, config-file TEI)
 
@@ -815,7 +813,7 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 - **Optimizer tests**:
   - FTS5-only: `thv vmcp serve --group default --optimizer` -> client sees only `find_tool`/`call_tool` -> `find_tool` discovers backend tools by keyword
   - Managed TEI: `thv vmcp serve --group default --optimizer-embedding` -> TEI container starts -> `find_tool` performs semantic search -> on shutdown, TEI container stops
-  - Fallback: TEI container fails to start -> vMCP falls back to FTS5-only with warning
+  - Fail-fast: TEI container fails to start -> `thv vmcp serve --optimizer-embedding` exits with a clear error
   - Idempotent: Running `thv vmcp serve --optimizer-embedding` twice reuses the existing TEI container
 - **Regression tests**: Verify that the standalone `vmcp serve` command still works identically after the refactor.
 - **Security tests**: Verify that quick mode binds to `127.0.0.1` only, that strict YAML parsing rejects unknown fields, and that HMAC session binding is enforced when configured.
@@ -826,6 +824,8 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 - **Architecture documentation**: `docs/arch/vmcp-local.md` covering local deployment, `docs/arch/vmcp-library.md` covering the library embedding pattern with brood-box as reference
 - **Examples**: `examples/vmcp-local-quickstart/` with a minimal setup, `examples/vmcp-advanced/` with auth, composite tools, and telemetry
 - **Existing docs updates**: Update `docs/arch/10-virtual-mcp-architecture.md` to reference the new CLI integration and library embedding path
+- **Docs website**: Update the ToolHive docs website to reflect the new `thv vmcp` subcommand and local optimizer story
+- **Legacy mcp-optimizer**: Once the native vMCP optimizer (Phase 4) is stable and documented, deprecate and archive the standalone `mcp-optimizer` Python project (`StacklokLabs/mcp-optimizer`). Remove references from the docs website and registry.
 
 ## Open Questions
 
@@ -841,9 +841,11 @@ The `thv vmcp` subcommand introduces no new attack surface beyond what the stand
 
 6. **TEI container lifecycle on crash**: If `thv vmcp serve` is killed ungracefully (SIGKILL, OOM), the TEI container will be left running. Should a cleanup mechanism be added (e.g., check for orphaned `thv-embedding-*` containers on startup)?
 
-7. **GPU support for TEI**: The default `cpu-latest` image works everywhere but is slower. Should `--optimizer-embedding` detect GPU availability and select a GPU-accelerated TEI image variant (e.g., `ghcr.io/huggingface/text-embeddings-inference:latest` for CUDA)?
+7. **GPU support for TEI**: The `--embedding-image` flag lets users select a GPU-accelerated TEI image variant (e.g., `ghcr.io/huggingface/text-embeddings-inference:turing-latest` for CUDA). Should `thv` also auto-detect GPU availability and suggest an appropriate image?
 
 8. **Embedding model recommendations**: Should `thv vmcp init` or docs recommend specific models for different use cases (small/fast vs. large/accurate)?
+
+9. **K8s dependency isolation (optional)**: The `thv` binary already transitively pulls `k8s.io/client-go` and `sigs.k8s.io/controller-runtime` through `pkg/container/kubernetes/`, so importing `pkg/vmcp/` does not introduce new module dependencies. However, the K8s-specific vMCP packages (`pkg/vmcp/k8s/`, `pkg/vmcp/workloads/k8s.go`) include code that is never activated in CLI mode. Gating these behind build tags or interfaces could reduce dead code in the `thv` binary, but this is a cleanup item rather than a blocker.
 
 ## References
 
--- a/rfcs/THV-XXXX-vmcp-local-experience.md
+++ b/rfcs/THV-XXXX-vmcp-local-experience.md
@@ -1,0 +1,875 @@
+# RFC-XXXX: First-Class vMCP Support in the Local CLI Experience
+
+- **Status**: Draft
+- **Author(s)**: Juan Antonio Osorio (@JAORMX)
+- **Created**: 2026-03-24
+- **Last Updated**: 2026-03-24
+- **Target Repository**: toolhive
+- **Related Issues**: N/A
+
+## Summary
+
+This RFC proposes integrating the Virtual MCP Server (vMCP) into the main `thv` CLI as a `thv vmcp` subcommand, elevating it from a standalone binary to a first-class feature of the local ToolHive experience. It also brings the vMCP optimizer — including managed lifecycle for the embedding service container — to the local developer workflow. Additionally, it formalizes the library usage path pioneered by brood-box, enabling downstream consumers to embed vMCP as a Go library for advanced integration scenarios.
+
+## Problem Statement
+
+vMCP is one of ToolHive's most powerful capabilities — it aggregates multiple backend MCP servers into a single unified endpoint, handles conflict resolution, supports composite tool workflows, and manages per-backend authentication. However, its current integration story has several gaps:
+
+- **Separate binary, separate experience**: vMCP ships as a standalone `vmcp` binary (`cmd/vmcp/`) with its own CLI, completely separate from the `thv` command users interact with daily. Users must discover, install, and manage a second binary. The `vmcp` binary is not included in goreleaser releases — users must build it manually or pull the container image.
+
+- **No discoverability**: Running `thv --help` gives no indication that vMCP exists. Users must read documentation to learn about the separate `vmcp` binary. This is at odds with ToolHive's goal of being a batteries-included MCP management platform.
+
+- **Inconsistent local experience**: `thv run` manages individual MCP servers, `thv group` organizes them, but there is no `thv` command to aggregate them. The logical workflow of "create group, add servers, aggregate via vMCP" breaks at the final step, requiring a context switch to a different binary.
+
+- **No official library usage guidance**: brood-box has successfully embedded vMCP as a Go library (using `pkg/vmcp/` packages directly), but this usage pattern is undocumented. Other downstream consumers who want to embed vMCP (e.g., IDE plugins, platform integrations) have no reference architecture to follow.
+
+- **Configuration fragmentation**: The standalone `vmcp` binary requires a full YAML config file even for simple local scenarios. Users who just want to aggregate their running group into a single endpoint must author a config file from scratch.
+
+- **Optimizer has no local story**: In Kubernetes, the vMCP optimizer is fully managed — the operator provisions an `EmbeddingServer` CRD that deploys the HuggingFace Text Embeddings Inference (TEI) container, wires the URL into the vMCP config, and gates readiness on the embedding service. Locally, users who want semantic tool search must manually `docker run` the TEI container and hand-wire the URL into a config file. There is no `thv` command to manage this lifecycle.
+
+All ToolHive users who manage more than one MCP server locally are affected. The lack of a unified CLI experience makes vMCP feel like an advanced/experimental feature rather than a core capability.
+
+## Goals
+
+- Add a `thv vmcp` subcommand with `serve`, `validate`, and `init` sub-commands that integrate vMCP into the main CLI
+- Support a zero-config quickstart: `thv vmcp serve --group <name>` should work without a config file for simple aggregation
+- Bring the optimizer to the local experience with managed embedding service lifecycle (`--optimizer` flag auto-manages a TEI container)
+- Maintain full feature parity with the standalone `vmcp` binary for users who need advanced configuration
+- Include vMCP in goreleaser so it ships with every ToolHive release (embedded in `thv` as a subcommand)
+- Document the library embedding pattern used by brood-box as an officially supported integration path
+- Ensure the standalone `vmcp` binary continues to work for backwards compatibility and Kubernetes deployments
+
+## Non-Goals
+
+- Removing the standalone `vmcp` binary (it remains the deployment target for Kubernetes via the operator)
+- Changes to the vMCP protocol, aggregation logic, or core architecture
+- Kubernetes-specific features (operator, CRD-based discovery, dynamic backend watching)
+- Integration with the long-running local server architecture (RFC-0034) — that is a natural follow-up but out of scope here
+- Desktop UI integration for vMCP management
+- Changes to composite tool workflows or the DAG execution engine
+
+## Proposed Solution
+
+### High-Level Design
+
+The `thv vmcp` subcommand wraps the existing vMCP server packages (`pkg/vmcp/`) behind a thin CLI layer, following ToolHive's established pattern where CLI commands in `cmd/thv/app/` are thin wrappers over `pkg/` logic.
+
+```mermaid
+flowchart TB
+    subgraph CLI["thv CLI (cmd/thv/)"]
+        Root["thv"]
+        VMCP["thv vmcp"]
+        Serve["thv vmcp serve"]
+        Validate["thv vmcp validate"]
+        Init["thv vmcp init"]
+    end
+
+    subgraph Standalone["Standalone Binary (cmd/vmcp/)"]
+        StdServe["vmcp serve"]
+        StdValidate["vmcp validate"]
+    end
+
+    subgraph Library["Library Consumers"]
+        BroodBox["brood-box"]
+        Custom["Custom Integrations"]
+    end
+
+    subgraph Core["pkg/vmcp/ (shared library)"]
+        Config["config/"]
+        Server["server/"]
+        Aggregator["aggregator/"]
+        Router["router/"]
+        Client["client/"]
+        Discovery["discovery/"]
+        Session["session/"]
+        Auth["auth/"]
+        Health["health/"]
+        Optimizer["optimizer/"]
+    end
+
+    subgraph Embedding["Embedding Service (optional)"]
+        TEI["TEI Container<br/>(managed or external)"]
+    end
+
+    Root --> VMCP
+    VMCP --> Serve
+    VMCP --> Validate
+    VMCP --> Init
+
+    Serve --> Core
+    Validate --> Core
+    StdServe --> Core
+    StdValidate --> Core
+    BroodBox --> Core
+    Custom --> Core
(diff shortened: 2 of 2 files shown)


rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+  - Quick mode (no config): `thv vmcp serve --group default` aggregates running servers
+- **Optimizer tests**:
+  - FTS5-only: `thv vmcp serve --group default --optimizer` -> client sees only `find_tool`/`call_tool` -> `find_tool` discovers backend tools by keyword
+  - Managed TEI: `thv vmcp serve --group default --optimizer-embedding` -> TEI container starts -> `find_tool` performs semantic search -> on shutdown, TEI container stops
+  - Fallback: TEI container fails to start -> vMCP falls back to FTS5-only with warning
+  - Idempotent: Running `thv vmcp serve --optimizer-embedding` twice reuses the existing TEI container
+- **Regression tests**: Verify that the standalone `vmcp serve` command still works identically after the refactor.
+- **Security tests**: Verify that quick mode binds to `127.0.0.1` only, that strict YAML parsing rejects unknown fields, and that HMAC session binding is enforced when configured.
+
+## Documentation
```
jerm-dro: The docs website will need to be updated. We can also then remove the legacy mcp-optimizer.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+
+- **Optimizer has no local story**: In Kubernetes, the vMCP optimizer is fully managed — the operator provisions an `EmbeddingServer` CRD that deploys the HuggingFace Text Embeddings Inference (TEI) container, wires the URL into the vMCP config, and gates readiness on the embedding service. Locally, users who want semantic tool search must manually `docker run` the TEI container and hand-wire the URL into a config file. There is no `thv` command to manage this lifecycle.
+
+All ToolHive users who manage more than one MCP server locally are affected. The lack of a unified CLI experience makes vMCP feel like an advanced/experimental feature rather than a core capability.
+
+## Goals
+
+- Add a `thv vmcp` subcommand with `serve`, `validate`, and `init` sub-commands that integrate vMCP into the main CLI
+- Support a zero-config quickstart: `thv vmcp serve --group <name>` should work without a config file for simple aggregation
+- Bring the optimizer to the local experience with managed embedding service lifecycle (`--optimizer` flag auto-manages a TEI container)
```
jerm-dro: You could more simply state the goal is to deprecate and later remove the legacy optimizer.

jerm-dro (commented):
LGTM! The fact this already works for broodbox indicates this is very feasible.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+- Maintain full feature parity with the standalone `vmcp` binary for users who need advanced configuration
+- Include vMCP in goreleaser so it ships with every ToolHive release (embedded in `thv` as a subcommand)
+- Document the library embedding pattern used by brood-box as an officially supported integration path
+- Ensure the standalone `vmcp` binary continues to work for backwards compatibility and Kubernetes deployments
+
+## Non-Goals
+
+- Removing the standalone `vmcp` binary (it remains the deployment target for Kubernetes via the operator)
```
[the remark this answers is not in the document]
JAORMX: Because it's a lot lighter weight to ship a container with just that binary.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+- **Security tests**: Verify that quick mode binds to `127.0.0.1` only, that strict YAML parsing rejects unknown fields, and that HMAC session binding is enforced when configured.
+
+## Documentation
+
+- **User documentation**: Quickstart guide for `thv vmcp` (zero-config and config-file modes), command reference
+- **Architecture documentation**: `docs/arch/vmcp-local.md` covering local deployment, `docs/arch/vmcp-library.md` covering the library embedding pattern with brood-box as reference
+- **Examples**: `examples/vmcp-local-quickstart/` with a minimal setup, `examples/vmcp-advanced/` with auth, composite tools, and telemetry
+- **Existing docs updates**: Update `docs/arch/10-virtual-mcp-architecture.md` to reference the new CLI integration and library embedding path
+
+## Open Questions
```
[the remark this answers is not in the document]
JAORMX: It's unclear so far. I haven't personally stumbled upon any issues, but we need to test and get more data.

aponcedeleonch (commented):
Hey! Really solid RFC, the whole local experience story makes a ton of sense. Left some inline comments — mostly questions and suggestions. Nada heavy, just thinking out loud. 🤙

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Good call. Added two explicit bullets in the Documentation section: one for the docs website update, and one for deprecating plus archiving `StacklokLabs/mcp-optimizer` once the native optimizer is stable (lines 827-828). Don't want to leave the Python project hanging once this lands.

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Done, much cleaner framing. The Goals section now states: "Deprecate and eventually remove the legacy standalone `mcp-optimizer` Python project (`StacklokLabs/mcp-optimizer`) once the native vMCP optimizer is stable" (line 37). Thanks!

rfcs/THV-0059-vmcp-local-experience.md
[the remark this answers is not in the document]
JAORMX: Revisiting this now that I've done more digging. Added Open Question #9 (line 848) with the current state: `pkg/container/kubernetes/` already transitively pulls `k8s.io/client-go` and `sigs.k8s.io/controller-runtime` into `thv`, so importing `pkg/vmcp/` doesn't introduce new module dependencies. The K8s-specific vMCP packages (`pkg/vmcp/k8s/`, `pkg/vmcp/workloads/k8s.go`) just sit there dormant in CLI mode. Gating them behind build tags is a nice cleanup but not a blocker.

Sharp edges around runtime behavior (e.g. discovery misdetecting K8s on a dev machine with a kubeconfig) are something I still want to test in anger once the prototype is running. The brood-box integration has been a good proof that this is doable, but I'd like more data before calling it a done deal.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+#   - name: "jira"
+#     url: "http://localhost:8081/mcp"
+#     transport: "streamable-http"
+
+aggregation:
+  conflictResolution: prefix
+  # tools:
+  #   - workload: "github"
+  #     filter: ["create_pr", "list_issues"]
+
```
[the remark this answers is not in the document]
JAORMX: You're right, this was a footgun waiting to happen. Killed the static-backends commented section from the `thv vmcp init` template and replaced it with a short note at lines 201-204:

```
# Backends are auto-discovered at runtime from the group.
# To modify backends, use `thv run` flags or group configuration
# rather than overriding them here.
```

Single source of truth, no drift between the vMCP config and the running group. Thanks for catching this!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+Implementation details:
+
+- **Container image**: `ghcr.io/huggingface/text-embeddings-inference:cpu-latest` (same image the K8s `EmbeddingServer` CRD uses)
+- **Container name**: `thv-embedding-<group>` (predictable, enables idempotent start/stop)
+- **Model cache**: Named volume `thv-embedding-model-cache` mounted at `/data` with `HF_HOME=/data`. This avoids re-downloading the model on every start (~130MB for `bge-small-en-v1.5`).
+- **Health check**: Poll `GET /health` with backoff until the TEI server reports ready. TEI must download and load the model on first start, which can take 30-60 seconds.
+- **Port binding**: Default `127.0.0.1:8384`. Chosen to avoid conflicting with the vMCP port (4483) or common dev ports.
+- **Lifecycle coupling**: The TEI container is started before the vMCP server and stopped after it shuts down. If the TEI container fails to start or becomes unhealthy, vMCP falls back to FTS5-only mode with a warning.
+- **Idempotent start**: If a `thv-embedding-<group>` container is already running (e.g., from a previous invocation), reuse it rather than creating a new one.
```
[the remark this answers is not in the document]
JAORMX: Great suggestion, added `--embedding-image` as a top-level flag. Now at line 141, with GPU examples at lines 415-421 (`text-embeddings-inference:turing-latest` for CUDA). Also tied this into the platform considerations bullet (line 457), which lets us sidestep GPU auto-detection entirely. That addresses Open Question #7 (line 844) too. Nice call linking those two threads.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+- **Platform considerations**: The TEI CPU image is amd64. On ARM64 hosts (Apple Silicon), the container runs under emulation. A future enhancement could detect architecture and select an appropriate image variant.
+
+**Tier 3: Full config control**
+
+Users with advanced requirements (custom embedding endpoint, GPU-accelerated TEI, external service) can configure the optimizer directly in the config file:
+
+```yaml
```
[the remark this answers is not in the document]
JAORMX: Good context, added it explicitly. Line 457 now reads:

> Platform considerations: The TEI CPU image is amd64-only. On ARM64 hosts (Apple Silicon), Docker Desktop handles this automatically via Rosetta 2 emulation, no `--platform` flag needed. The overhead is roughly 5-15% for CPU-bound workloads, which is acceptable for embedding generation.

And as you noted, `--embedding-image` lets users opt into architecture-native images when available. The 5-15% number is useful, saves people from chasing weird perf reports. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+- **Container name**: `thv-embedding-<group>` (predictable, enables idempotent start/stop)
+- **Model cache**: Named volume `thv-embedding-model-cache` mounted at `/data` with `HF_HOME=/data`. This avoids re-downloading the model on every start (~130MB for `bge-small-en-v1.5`).
+- **Health check**: Poll `GET /health` with backoff until the TEI server reports ready. TEI must download and load the model on first start, which can take 30-60 seconds.
+- **Port binding**: Default `127.0.0.1:8384`. Chosen to avoid conflicting with the vMCP port (4483) or common dev ports.
+- **Lifecycle coupling**: The TEI container is started before the vMCP server and stopped after it shuts down. If the TEI container fails to start or becomes unhealthy, vMCP falls back to FTS5-only mode with a warning.
+- **Idempotent start**: If a `thv-embedding-<group>` container is already running (e.g., from a previous invocation), reuse it rather than creating a new one.
+- **Platform considerations**: The TEI CPU image is amd64. On ARM64 hosts (Apple Silicon), the container runs under emulation. A future enhancement could detect architecture and select an appropriate image variant.
+
```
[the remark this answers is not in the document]
JAORMX: Agreed, per-group coupling was the wrong default. Swapped the naming scheme to `thv-embedding-<model-short-hash>` (line 451). Multiple vMCPs using the same embedding model now share one TEI container, and different models get their own. Matches how you'd expect it to behave without having to think about it. Good catch!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+**Tier 3: Full config control**
+
```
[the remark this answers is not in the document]
JAORMX: Yep, fixed port was a bad default. Line 454 now uses `pkg/networking.FindAvailable()`, the same mechanism `thv run` uses when no port is given. The allocated port gets logged so the user can see it. No specific reason for 8384, it was just convenient for the config-file example, but reusing the existing pattern is cleaner. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+Users with advanced requirements (custom embedding endpoint, GPU-accelerated TEI, external service) can configure the optimizer directly in the config file:
```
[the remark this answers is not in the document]
JAORMX: You're right, and I went back and forth on this one. Landed on fail-fast, exactly as you argued. Line 455:

> If the TEI container fails to start or become healthy when `--optimizer-embedding` was explicitly requested, `thv vmcp serve` exits with a clear error message... silently degrading to FTS5-only would mask environment issues (Docker not running, port conflicts, etc.) and produce confusingly poor results.

Also echoed in Security Considerations > Mitigations (line 719). If someone wants keyword-only, they pass `--optimizer` explicitly. Thanks for pushing back on the original framing!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+    Model    string // default: BAAI/bge-small-en-v1.5
+    Port     int    // default: 8384
+    GroupRef string // used in container name: thv-embedding-<group>
+}
+
+// EmbeddingServiceManager manages the lifecycle of a local TEI container.
+type EmbeddingServiceManager struct { ... }
+
+// Start launches (or reuses) the TEI container and waits for readiness.
+// Returns the embedding service URL.
```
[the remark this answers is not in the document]
JAORMX: Labels all the way, you convinced me. Line 513 now describes ownership via `pkg/labels/`, specifically the `toolhive` label (ToolHive-managed) and `toolhive-auxiliary` (marks it as an auxiliary workload, same pattern the inspector container uses). On startup we check for a running container with matching name + labels: if it's there and healthy, reuse it. If a container with the expected name is running without the `toolhive` label, we assume the user is managing their own TEI and don't touch it. No extra config surface, no state drift. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,875 @@
...
+
+- **Local binding**: `thv vmcp serve` defaults to `127.0.0.1`, preventing remote access. Users must explicitly set `--host 0.0.0.0` to expose the endpoint.
+- **Incoming auth**: Supports OIDC and anonymous modes via config. Quick mode defaults to anonymous, which is appropriate for local single-user scenarios.
+- **Outgoing auth**: Per-backend authentication strategies (header injection, token exchange, unauthenticated) configured via the config file. Quick mode defaults to unauthenticated.
+- **Cedar authorization**: When configured, Cedar policies gate access to individual tools and resources. The library embedding path (brood-box) demonstrates this with `observe`, `safe-tools`, and `full-access` profiles.
+- **No privilege escalation**: The `thv vmcp` subcommand runs with the same privileges as the invoking user. No root or elevated permissions required.
+
+### Data Security
+
+- vMCP proxies MCP protocol messages between clients and backends. It does not persist tool call results or resource contents.
```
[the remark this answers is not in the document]
JAORMX: Good food for thought, and I think you're right that it sidesteps the auth/port-binding question cleanly for local single-user. Added it as a future direction under Security Considerations > Authentication and Authorization (line 682):

> Unix socket transport (future): For local single-user scenarios, a Unix socket transport (similar to `thv serve --socket`) would sidestep port-binding and auth concerns entirely, relying on filesystem permissions for access control. This is a natural follow-up but out of scope for this RFC.

Keeps the UX aligned with where `thv serve` is heading. Out of scope for this RFC so we can focus on landing the vMCP integration first, but definitely on the roadmap.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+
+```go
+var informationalCommands = map[string]bool{
+    "version":    true,
+    "search":     true,
+    "completion": true,
+    "registry":   true,
+    "mcp":        true,
+    "skill":      true,
+    "vmcp":       true, // new
```
yrobla, line 279: **Contradiction: `vmcp` marked informational but `--optimizer-embedding` requires the container runtime**

The justification given for adding `vmcp` to `informationalCommands` is that "vMCP is a proxy that connects to already-running backends". That's true for `thv vmcp validate` and `thv vmcp init`, but `thv vmcp serve --optimizer-embedding` explicitly needs the container runtime to launch the managed TEI container — the very runtime the informational check is skipping.

This means a user with no container runtime who runs `thv vmcp serve --optimizer-embedding` will get a confusing late failure (can't start TEI container) instead of the clear early error the runtime check is supposed to provide.

**Suggested fix:** Don't add `vmcp` to `informationalCommands` as a blanket entry. Instead, handle the check at the serve subcommand level: skip the runtime check unless `--optimizer-embedding` is passed. `validate` and `init` can be registered individually as informational if that mechanism supports per-subcommand registration, or the check can be deferred to inside the `serve` handler.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+- **Port binding**: A random available port is allocated using ToolHive's existing `pkg/networking.FindAvailable()` pattern (the same mechanism `thv run` uses when no explicit port is given). The allocated port is reported in the logs. This avoids conflicts when multiple vMCP instances or other services are running.
+- **Lifecycle coupling**: The TEI container is started before the vMCP server and stopped after it shuts down. If the TEI container fails to start or become healthy when `--optimizer-embedding` was explicitly requested, `thv vmcp serve` exits with a clear error message. The `--optimizer-embedding` flag is a clear signal that the user wants semantic search — silently degrading to FTS5-only would mask environment issues (Docker not running, port conflicts, etc.) and produce confusingly poor results. Users who want keyword-only search should use `--optimizer` explicitly.
+- **Idempotent start**: If a TEI container with the matching name and ToolHive labels is already running (e.g., from a previous invocation or another vMCP instance with the same model), reuse it rather than creating a new one.
```
yrobla, line 456: **TEI container sharing has an unaddressed reference-counting gap**

The idempotent-start behavior (reuse a running container if name + labels match) is a good idea for efficiency. But combined with the lifecycle-coupling behavior (stop TEI container on shutdown), this creates a race:

1. Instance A (`--group default`) starts a TEI container named `thv-embedding-bge-sm-a1b2c3`.
2. Instance B (`--group engineering`, same model) reuses that container.
3. Instance A shuts down gracefully → stops the TEI container.
4. Instance B's semantic search now silently fails or returns errors.

The section says "The TEI container is started before the vMCP server and stopped after it shuts down", but it doesn't say what happens when the container is shared.

**Suggested options (pick one):**
- **Label-based ref counting**: increment/decrement a `thv-embedding-refcount` label on the container. Only stop when count reaches 0.
- **No cleanup on shared containers**: only stop the TEI container if the current process was the one that started it (track via an instance label). Shared containers are left running; document that users can `docker stop` them.
- **No sharing**: use a unique container name per vMCP instance (e.g., include a random suffix) — simpler lifecycle, slightly more resource usage.

The simplest option is probably no-sharing with a per-instance name. The "avoid duplication" benefit of sharing is marginal for local dev use.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+Starts a local vMCP server that aggregates MCP servers from a ToolHive group.
+
+```
+thv vmcp serve [flags]
+
+Flags:
+  -c, --config string            Path to vMCP configuration file (optional)
+  -g, --group string             ToolHive group to aggregate (default: "default")
+      --host string              Host to bind to (default: "127.0.0.1")
+  -p, --port int                 Port to listen on (default: 4483)
```
yrobla, line 136: **Port 4483 collides for multiple concurrent `thv vmcp serve` instances**

The RFC correctly uses `FindAvailable()` for the TEI container port, but the vMCP server itself defaults to a fixed port 4483. If a user has two groups running simultaneously — e.g., `thv vmcp serve --group default` and `thv vmcp serve --group engineering` — both default to 4483 and the second will fail to bind.

This is worth addressing explicitly since the "zero-config quickstart" story implies users will run this casually, without thinking about port management.

**Suggestion:** Either (a) auto-assign a random port when 4483 is taken (log the chosen port clearly), or (b) document prominently that users must specify `-p` for concurrent instances. Option (a) is friendlier for the quickstart UX but means the client config URL changes on each invocation — which matters for IDE config files (Open Question 2). The RFC should take a position.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+      --embedding-image string   TEI container image (default: "ghcr.io/huggingface/text-embeddings-inference:cpu-latest")
+```
+
+Two modes of operation:
+
+1. **Quick mode** (no config file): When `--config` is omitted, the command auto-generates a minimal configuration using `--group` as the `groupRef`. This provides a zero-config quickstart for the common case of "aggregate everything in my group."
+
+2. **Config mode**: When `--config` is provided, the command loads the full YAML configuration, supporting all advanced features (aggregation policies, composite tools, auth, telemetry, etc.). This is identical to the standalone `vmcp serve` behavior.
+
+Note that `groupRef` is a **required field** in the vMCP config (see `pkg/vmcp/config/config.go`). In quick mode, `--group` populates this field in the auto-generated config. In config mode, the group comes from the config file's `groupRef` field. If both `--config` and `--group` are provided, the CLI flag overrides the config file's `groupRef`, which is useful for reusing a config template across multiple groups.
```
yrobla, line 150: **`--group` overriding `groupRef` in a provided config file is a footgun**

Allowing `--group` to silently override `groupRef` when `--config` is also provided is surprising. A config file is typically authored with a specific group in mind — it may have auth configurations, composite tool definitions, or backend-specific aggregation policies that only make sense for that group. Substituting a different `groupRef` while keeping the rest of the config intact could produce subtly wrong behavior (e.g., using auth credentials designed for group A's backends when talking to group B's backends).

**Suggestion:** Emit a warning (or require an explicit `--force-group` flag) when both `--config` and `--group` are provided. Alternatively, restrict the quick-mode `--group` flag to only work when `--config` is absent, and document that group selection with a config file should be done by editing `groupRef` in the file. The "template reuse" use case can still be served by `thv vmcp init`, which generates per-group configs.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+
+2. **Should `thv vmcp serve` register with client configuration files?** When the long-running server starts, it writes to IDE config files. Should `thv vmcp serve` do the same (write the vMCP endpoint into Claude Code, VS Code, etc. config files)?
+
+3. **Should we add `thv vmcp status`?** A command to query the health and capabilities of a running vMCP instance (list aggregated tools, backend health) using the existing `/status` and `/api/backends/health` endpoints.
+
+4. **Stability level annotations for library packages**: The stability table in this RFC assigns levels based on current usage patterns. Should these be formalized with `doc.go` annotations in each package?
+
+5. **Quick mode auth defaults**: Quick mode defaults to `anonymous` incoming auth and `unauthenticated` outgoing auth. Is this appropriate, or should it attempt to detect backend auth requirements from the group's workload configurations?
+
+6. **TEI container lifecycle on crash**: If `thv vmcp serve` is killed ungracefully (SIGKILL, OOM), the TEI container will be left running. Should a cleanup mechanism be added (e.g., check for orphaned `thv-embedding-*` containers on startup)?
```
yrobla, line 842: **Open Question 6 (orphan TEI containers) should be resolved in this RFC, not deferred**

This is framed as a question but the answer is clearly "yes, we need cleanup" — orphaned TEI containers from SIGKILL or OOM will accumulate silently, each consuming ~500MB-1GB of memory for the model weights.

Leaving this unresolved means Phase 4 implementers will have to design and decide this under time pressure. The RFC is the right place to settle it.

**Proposed resolution to include in the RFC:** On `thv vmcp serve` startup, scan for containers with the `toolhive-auxiliary` label and a `thv-embedding-*` name that are not currently referenced by a running vMCP instance. Report them to the user (`thv: found orphaned TEI container from previous session: thv-embedding-bge-sm-a1b2c3. Run 'thv vmcp cleanup' to remove it.`). Add `thv vmcp cleanup` as a new subcommand (Phase 4 scope) that removes orphaned embedding containers and optionally the model cache volume. This pairs naturally with the idempotent-start behavior.

yrobla (commented):
This is a well-structured RFC with a clear problem statement, a sensible three-tier optimizer design, and good backward-compatibility story. The shared `pkg/vmcp/cli/` package is the right architectural call — it prevents logic duplication between `thv vmcp` and `cmd/vmcp/`. The library embedding section is also a welcome addition, since the brood-box pattern has been undocumented for a while.

I have five concrete issues to raise, in rough priority order. None are blockers for Draft status but a few (especially the `informationalCommands` contradiction and the TEI sharing gap) should be resolved before implementation starts.

aponcedeleonch (approved):
nice

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+
+Flags:
+  -g, --group string    ToolHive group to generate config for (default: "default")
+  -o, --output string   Output file path (default: stdout)
+      --conflict-resolution string   Conflict resolution strategy: prefix|priority (default: "prefix")
+```
+
+This command:
+1. Discovers all running workloads in the specified group via `groups.NewCLIManager()` and `workloads.NewManager()`
+2. Generates a YAML config with backends pre-populated from the discovered workloads
```
yrobla, line 190: **`thv vmcp init` behavior when no workloads are running is unspecified**

The command discovers "all running workloads in the specified group". But what if the user runs `thv vmcp init --group engineering` before they've started any MCP servers? This is a plausible scenario — scaffolding the config before starting servers, or doing so in a CI environment without running containers.

The RFC should specify: does `init` fail if no workloads are found? Does it generate an empty but valid config with a comment explaining that backends are discovered at runtime? Does it warn the user?

**Suggested behavior:** `init` should succeed even with zero running workloads, generating a valid config with an empty backends list and a comment: `# No running workloads found in group 'engineering'. Start servers with 'thv run' and re-run 'thv vmcp init' to populate backends.` This makes the command useful for config templating without requiring a live environment.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+    "completion": true,
+    "registry":   true,
+    "mcp":        true,
+    "skill":      true,
+    "vmcp":       true, // new
+}
+```
+
+**`pkg/vmcp/cli/`** (new package)
+
```
yrobla, line 284: **`pkg/vmcp/cli/` mixes CLI-assembly logic with infrastructure; `EmbeddingServiceManager` belongs in `pkg/vmcp/embedding/`**

The shared serve/validate/init logic in `pkg/vmcp/cli/` makes sense: two `cmd/` packages need to share it, and `pkg/` is the idiomatic Go location for shared library code. But placing the `EmbeddingServiceManager` in the same `cli` package creates a layering problem.

The `EmbeddingServiceManager` is container infrastructure — it pulls an image, starts a container, polls health, manages a named volume. This capability is useful independently of any CLI: a test harness, a desktop app, or a library consumer might want to programmatically start a managed TEI container without going through the CLI layer.

Putting it in `pkg/vmcp/cli/` signals "don't import this outside CLI code", which is wrong. The existing library embedding guidance tells consumers to only import `pkg/vmcp/*` packages. If they ever want a managed embedding lifecycle, they'd have to go through the `cli` package, which is surprising.

**Suggested split:**
- `pkg/vmcp/embedding/` — `EmbeddingServiceManager`, `EmbeddingServiceConfig`, container lifecycle management. Clean dependency: only `pkg/container/` and `pkg/labels/`.
- `pkg/vmcp/cli/` — flag-to-config assembly (`ServeConfig` from cobra flags), quick-mode config generation, `Serve()`, `Validate()`, `Init()`. Clean dependency: `pkg/vmcp/embedding/` and `pkg/vmcp/`.

This also prepares the ground for exposing `EmbeddingServiceManager` to library consumers (your stability table could add `pkg/vmcp/embedding` as Stable once proven).

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+##### Background: How the Optimizer Works
+
+The vMCP optimizer is a context-window reduction system. When enabled, instead of exposing all aggregated tools to the LLM client (which can consume enormous token budgets), vMCP exposes only two meta-tools:
+
+- **`find_tool`** — Semantic and keyword search over the full tool set
+- **`call_tool`** — Dynamic invocation of any backend tool by name
+
+The optimizer uses an in-memory SQLite FTS5 index for keyword search (BM25 ranking) and optionally an external embedding service for semantic vector search. The SQLite store runs inside the vMCP process — no additional infrastructure is needed for keyword-only mode. For hybrid search (keyword + semantic), an external HuggingFace Text Embeddings Inference (TEI) service provides vector embeddings.
+
```
yrobla, line 341: **Session-scoped FTS5 index rebuild latency is not addressed for local use**

THV-0022 (referenced at the end of the RFC) describes the optimizer as having a session-scoped SQLite index: the FTS5 index is built per MCP client session, from scratch, by indexing all aggregated tool descriptions at session init time.

For a local developer with many backends (say 5 servers × 10-20 tools each = 50-100 tools), this index rebuild happens every time a new conversation starts in Claude Code or another client. For keyword-only FTS5 it's fast. For semantic search (`--optimizer-embedding`), it means embedding 50-100 tool descriptions via HTTP calls to the TEI service at the start of every session — potentially seconds of latency before the first tool call succeeds.

The RFC doesn't acknowledge this tradeoff or propose any mitigation (e.g., persistent index across sessions, warm-up on server start rather than on client connect, pre-indexing at vMCP startup). This is going to be one of the first UX complaints after launch for users with more than ~30 tools.

**The RFC should address this:** Either (a) confirm that session-scoped indexing is acceptable for the local use case and explain why, (b) propose that the local vMCP pre-warms the index at server startup so the first client connection doesn't pay the full indexing cost, or (c) add this as an explicit open question.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+- **Container name**: `thv-embedding-<model-short-hash>` (e.g., `thv-embedding-bge-sm-a1b2c3`), derived from a hash of the model name. This allows multiple vMCP instances using the same embedding model to share a single TEI container, avoiding unnecessary duplication. Different models get separate containers.
+- **Model cache**: Named volume `thv-embedding-model-cache` mounted at `/data` with `HF_HOME=/data`. This avoids re-downloading the model on every start (~130MB for `bge-small-en-v1.5`).
+- **Health check**: Poll `GET /health` with backoff until the TEI server reports ready. TEI must download and load the model on first start, which can take 30-60 seconds.
+- **Port binding**: A random available port is allocated using ToolHive's existing `pkg/networking.FindAvailable()` pattern (the same mechanism `thv run` uses when no explicit port is given). The allocated port is reported in the logs. This avoids conflicts when multiple vMCP instances or other services are running.
```
yrobla, line 454: **First-run UX gap: image pull time is omitted; only model download time is mentioned**

The RFC mentions "TEI must download and load the model on first start, which can take 30-60 seconds" and that the model is cached in a named volume. Good. But there's a prior step that can take much longer: pulling the TEI container image itself.

`ghcr.io/huggingface/text-embeddings-inference:cpu-latest` is several GB. On a fresh machine or after a Docker image prune, the pull can take several minutes on a slow connection. The RFC says nothing about what the user sees during this time.

**Two gaps to address:**
1. **Image pull progress**: Should the CLI print `Pulling TEI image (this may take a few minutes on first run)...` before starting the pull? Or stream Docker pull progress? Without this, `thv vmcp serve` will appear to hang.
2. **Pull vs. model download distinction**: After the pull, the 30-60 second model download happens inside the container. The RFC should describe the full first-run timeline: ~image pull (minutes, network-dependent) → container start → model download (30-60s) → health ready → vMCP starts. Each phase needs user-visible feedback.

This is a quickstart experience RFC — the first-run UX is the most important moment to get right.

rfcs/THV-0059-vmcp-local-experience.md
```diff
@@ -0,0 +1,877 @@
...
+```
+
+#### Stability Guarantees
+
+The following `pkg/vmcp/` packages are considered stable for library consumers:
+
+| Package | Stability | Notes |
+|---------|-----------|-------|
+| `pkg/vmcp` (root) | Stable | `BackendRegistry`, `ImmutableRegistry`, core types |
+| `pkg/vmcp/config` | Stable | Config loading and validation |
```
yrobla, line 636: **Stability table needs a versioning/release policy to be meaningful**

The RFC declares packages as Stable/Beta/Internal but doesn't specify what "Stable" means in terms of toolhive's release process. Specifically:

- Does "Stable" mean no breaking changes between minor releases? Major releases? Ever?
- Is toolhive on a semver v1+ scheme where `pkg/vmcp/` is part of the public API contract?
- How will downstream consumers (brood-box and future ones) be notified of deprecations?
- What's the promotion path from Beta → Stable (e.g., two release cycles without changes, explicit RFC to promote)?

Without answers to these, "officially supported integration path" is a social contract with no enforcement mechanism. brood-box has already pinned to specific toolhive commits to manage this risk — formalizing the path means actually committing to a notification process for breaking changes.

**Suggested addition:** A brief subsection under Library Embedding Pattern: `### Versioning Policy` that states the stability guarantee in terms of toolhive's release tags, and defines the process for deprecating a Stable API (e.g., deprecated in release N, removed in release N+2 with a migration guide). Even a minimal policy is better than none.

yrobla (commented):
## Deeper Review: Motivation, Concept, and Architecture

Building on my [previous review](https://github.com/stacklok/toolhive-rfcs/pull/59#pullrequestreview-4168657980) (which covered five implementation-level issues), this review looks at the broader picture: whether the motivation holds up, whether the concept is the right one, and where the design has gaps that will create friction.

---

### Motivation: Mostly solid, with one overstatement

The core problem statement is accurate and well-articulated. vMCP is genuinely invisible to `thv` users today, the config-file requirement creates friction for simple use cases, and the local optimizer story is a real gap. These are worth solving.

The one overstatement: _"All ToolHive users who manage more than one MCP server locally are affected."_ The optimizer and composite tool features are only relevant to a subset of users. Most users with two or three MCP servers probably don't need tool aggregation at all — they just pick tools from individual servers manually. The stronger case is for users with 5+ servers where tool discovery becomes unwieldy. Worth calibrating the scope of who this is actually for, since it affects prioritization of tiers and the zero-config story.

---

### Concept: `thv vmcp` is the right namespace — but the command verbs need work

I agree with the RFC's rejection of Alternative 3 (`thv group serve`): vMCP is a distinct concept with its own lifecycle, and grouping it under `group` would hide the advanced features. The `thv vmcp` namespace is correct.

However, `thv vmcp serve` as the zero-config quickstart is ergonomically awkward. `vmcp` is an implementation term — users think in terms of "aggregate my tools" or "unify my MCP servers", not "start a virtual MCP server". The RFC's goal is to make this a first-class experience, but naming it after an internal protocol concept creates a mismatch between user intent and command vocabulary. This is worth a deliberate UX decision, even if the conclusion is "keep vmcp" for consistency with the existing binary.

---

### Architecture: Three areas need rethinking

**1. `pkg/vmcp/cli/` is a poor home for the `EmbeddingServiceManager`**

The shared serve logic being in `pkg/vmcp/cli/` is justifiable (it's reused by two `cmd/` packages). But the `EmbeddingServiceManager` — which manages a container lifecycle — does not belong in a `cli` package. It is infrastructure, not CLI logic. Downstream library consumers (the very audience the library embedding section addresses) might want to embed this capability without any CLI dependency. Putting it in `pkg/vmcp/cli/` signals "this is CLI plumbing, don't import it", which is the wrong signal.

Better split: `pkg/vmcp/embedding/` for the `EmbeddingServiceManager` (callable by anyone who wants a managed TEI lifecycle), `pkg/vmcp/cli/` for the config-from-flags assembly logic that only CLI callers need.

**2. The optimizer UX is underspecified for local use**

See inline comments (lines 341 and 454) for the two specific gaps: session-scoped index rebuild latency and first-run image pull time. These will be the two most common UX complaints after launch, and neither is addressed. The RFC references THV-0022 for the optimizer architecture but doesn't carry the local-UX implications of session-scoped indexing forward.

**3. Library embedding: the stability table is declared but not enforced**

See inline comment (line 636). "Officially supported" needs a versioning policy to be meaningful. Without one, this section creates expectations the project can't reliably honor.

---

### Alternatives worth reconsidering (or addressing more directly)

**Should `thv vmcp serve` also auto-register with client configs?**

Open Question 2 asks whether `thv vmcp serve` should write the endpoint into Claude Code / VS Code config files. This is actually the most important UX question in the RFC and it's buried as an open question. If the answer is yes, it dramatically changes the scope of this RFC (and the port-stability question from my previous review becomes critical — a fixed port 4483 becomes load-bearing infrastructure). If the answer is no (for now), say so explicitly and explain the tradeoff.

**Should zero-config quickstart be `thv group serve` as a thin alias?**

Alternative 3 is dismissed but a thin alias `thv group serve <name>` → `thv vmcp serve --group <name>` would dramatically improve discoverability for the simple case. Users discovering groups via `thv group ls` would naturally try `thv group serve`. The two approaches aren't mutually exclusive — `thv vmcp` can have the full feature set while `thv group serve` provides the most common entry point. This is worth a sentence in the RFC even if rejected.

---

### Structural simplification opportunity

Phases 1-3 (extract logic, add subcommands, quick mode) are implementation work. Phase 4 (optimizer with managed TEI) is a substantially larger feature with its own UX, lifecycle, and operational surface area. The RFC would be stronger — and easier to implement incrementally — if Phase 4 had its own RFC (or at minimum a clear "Phase 4 is a stretch goal" signal). As written, the detailed TEI design is given equal weight to the core `thv vmcp` subcommand, which makes the RFC harder to approve because reviewers must evaluate two different scope/risk levels at once.

---

### Summary of new inline comments below

1. **Line 190** — `thv vmcp init` behavior when no workloads are running is unspecified
2. **Line 284** — `pkg/vmcp/cli/` mixes CLI and infrastructure concerns; `EmbeddingServiceManager` should be in `pkg/vmcp/embedding/`
3. **Line 341** — Session-scoped FTS5 index rebuild latency is not addressed for local use
4. **Line 454** — First-run UX: image pull time is omitted; only model download time is mentioned
5. **Line 636** — Stability table needs an explicit versioning/release policy to be meaningful

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Yeah, you're right. The blanket entry hides exactly the failure mode the runtime check was designed to surface. So I split it up: `validate` and `init` stay informational (no runtime calls), and `serve` does the check inside its handler, required when `--optimizer-embedding` is set, skipped otherwise. A user without Docker who asks for the managed TEI container now hits an early, explicit error instead of a confusing late one. Lines 269-289 have the new wording. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Good catch, this is a real race and I didn't call it out properly. Went with the "owner-only cleanup" option: each vMCP instance tags the container it *creates* with a `toolhive-embedding-owner=<instance-id>` label, and only that instance stops it on exit. Instances that reused an existing container via idempotent-start leave it alone. The trade-off is we can leave containers running across sessions, but that's fine as long as we give users a way to clean them up, which we're now doing via `thv vmcp cleanup` (see my reply to your Open Question 6 comment). Updated lines 456-462. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Good point, the quickstart story does imply casual usage and we should take a position. Landed on "fixed 4483, fail-fast with an actionable error" rather than auto-randomizing. Two reasons: most users run one vMCP per box and a predictable URL keeps the client-config story simple (ties into your Open Question 2 point too), and auto-randomizing would break every quickstart guide and every IDE config snippet we ask people to paste. When the port is already taken we exit with `port 4483 already in use, try 'thv vmcp serve -p 0' to pick a free port` so the power-user case (two groups side by side) is still one flag away. New "Port binding policy" subsection around line 167. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Agreed, the "useful for reusing a config template" framing was me rationalizing a footgun. Switched to "error if both are provided" rather than trying to warn. A config file often carries auth and aggregation policies that only make sense for a specific group, and swapping `groupRef` while leaving the rest alone is exactly the kind of subtle bug we want to avoid. The "template reuse" story is already covered by `thv vmcp init`, which generates a per-group config the user can then edit. Error message wording and justification are at lines 150-161. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Agreed, this is the right call. Promoted it out of Open Questions and into the body as a new `thv vmcp cleanup` subcommand with a supporting startup scan in `thv vmcp serve`. The mechanics: containers carry a `toolhive-embedding-owner=<instance-id>` label, `cleanup` enumerates `thv-embedding-*` containers and removes any whose owner is no longer alive, and the serve startup emits a one-line hint when it detects orphans (`thv: found N orphaned embedding container(s)... run 'thv vmcp cleanup'`). `--model-cache` and `--dry-run` flags are there too. This pairs with the owner-only cleanup story from your line-456 comment. See the new "Orphan cleanup: `thv vmcp cleanup`" subsection, and Open Question 6 is gone. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Good point, I had this case in my head but never wrote it down. Spec'd it explicitly: `thv vmcp init --group <name>` succeeds even when the group is empty or doesn't exist yet, emits a valid config with empty backends and a comment pointing at `thv run --group ...` as the next step, and prints a single stderr warning so the result isn't silent. That keeps `init` useful for config templating in CI without pretending everything's fine. Added right after the `init` bullets. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Yeah, you're right. Putting infrastructure in a package named `cli` was a layering smell I wrote into the RFC without thinking. Split into two packages:

- `pkg/vmcp/embedding/` — container lifecycle (pull, start, health, owner-only cleanup, orphan enumeration). Depends only on `pkg/container/` and `pkg/labels/`. Usable by test harnesses, desktop apps, and future library consumers without dragging in cobra.
- `pkg/vmcp/cli/` — CLI-assembly only: flag translation, quick-mode config generation, `Serve()`/`Validate()`/`Init()`. Depends on `pkg/vmcp/embedding/` and `pkg/vmcp/`.

Also added both packages to the stability table (`embedding` = Beta until Phase 4 lands, `cli` = Internal) and updated the Phase 4 implementation plan accordingly. Lines 338 (cli scoping note), 528 onward (new embedding section), and the stability table. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Great catch, this would 100% have been the first UX complaint after launch. The session-scoped indexing model in THV-0022 is fine for keyword-only (FTS5 over 100 tools is sub-second) but it's rough for semantic search, where every new Claude Code conversation would pay "embed 50-100 tool descriptions" before the first `find_tool` works.

Resolved in the RFC by pre-warming at server start: `thv vmcp serve --optimizer-embedding` builds the FTS5 index and populates an embedding cache keyed on `(model, tool-description-hash)` once, at boot, and gates `/readyz` on it. Sessions still get their own logical view, but the embeddings themselves are pure functions of the input so they're cached across sessions. Per-session cost drops from "embed N tools" to "look up N hashes." Added a new "Index warm-up for the local case" subsection with the trade-offs (slower startup, tiny memory overhead, cache invalidation on model swap). Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: You're right, I conflated "model download" with the full first-run story. Spelled out the actual timeline in the RFC: image pull (several GB, minutes on a slow link, first run only) → container start → model download (30-60s, once per volume) → index warm-up (a few seconds, from your other comment) → vMCP ready. Added user-visible progress lines for each phase so `thv vmcp serve --optimizer-embedding` doesn't look hung on first run:

```
thv: pulling embedding image 'ghcr.io/huggingface/text-embeddings-inference:cpu-latest'
     (first run only, several GB, this may take a few minutes)...
thv: image ready
thv: starting embedding container 'thv-embedding-bge-sm-a1b2c3' on 127.0.0.1:47231
thv: downloading model 'BAAI/bge-small-en-v1.5'...
thv: model ready
thv: warming optimizer index (50 tools)... done in 2.8s
thv: vMCP listening on http://127.0.0.1:4483/mcp
```

Subsequent runs skip pull and model download, only warm-up runs. Lines 468-480. Thanks!

rfcs/THV-0059-vmcp-local-experience.md
JAORMX: Good point, a stability label without a release process attached is a social contract with no enforcement. Added a "Versioning policy" subsection under Library Embedding Pattern that defines what each level means in practice:

- **Stable**: no breaking changes inside a minor release; breaking changes at minor boundaries only after a two-release deprecation cycle (deprecated in `N`, removed no earlier than `N+2`).
- **Beta**: may change between minor releases, one-release deprecation cycle; promoted to Stable after two consecutive minor releases with no breaks and at least one external consumer.
- **Internal**: no guarantees.

Plus an explicit notification pipeline: breaking changes and Beta→Stable promotions get a "Library API Changes" section in release notes, known downstream consumers (brood-box today) are tagged on the release PR, and the CHANGELOG links to a migration snippet per break. Intentionally light, just enough for brood-box and the next consumer to trust the table. Thanks!

JAORMX:
@yrobla thanks for the deeper pass, these five were sharper than the inline ones and worth addressing directly in the RFC. Working through them:

**1. Motivation overstatement.** Fair call, this was me writing the Problem Statement with too much confidence. Narrowed the last paragraph (line 30) to say severity scales with the number of servers: 2-3 you pick tools manually and barely notice, 5+ is where tool discovery and context-window cost start hurting. That lines up with the "50-100 tools from 5-10 backends" threshold the optimizer section already uses around line 398.

**2. `thv vmcp` as an acronym instead of a verb.** You're right, this breaks the verb pattern the rest of the CLI follows (`run`, `build`, `proxy`, `serve`, `group`), and I'd been treating that as obvious rather than justifying it. Rewrote "Why not chosen" in Alternative 3 to call it out: we're picking the acronym on purpose because hiding vMCP behind an intent-verb makes composite tools, auth, and optimizer tiers harder to find, and because splitting from the existing standalone `vmcp` binary (different name for the same thing) would be worse than an acronym. The user-intent framing lives in the Short description (`cmd/thv/app/vmcp.go`), which is what shows up in `thv --help`.

**3. Open Question 2 (client config auto-registration) is buried.** Yeah, and looking at it again the answer is clearly yes. `thv run` already does this today via `pkg/client/manager.go` (`AddServerToClients` on start, `RemoveServerFromClients` on stop). Not doing the same for `thv vmcp serve` would be an inconsistency, not a feature. Pulled this out of Open Questions and into the main design as a new "Client configuration registration" subsection after the Port binding policy. On startup we register a `vmcp-<group>` entry pointing at `http://127.0.0.1:4483/mcp`, on graceful shutdown we remove it. The fixed-port policy from your previous review is what keeps the registered URL stable, so the two pieces fit together. Added a `--no-client-register` escape hatch for users behind reverse proxies or managing IDE configs manually. Open Question 2 is gone.

**4. `thv group serve` as a thin alias.** Good point, the two approaches aren't mutually exclusive. Added to Alternative 3 as a "Future direction": not doing it in this RFC because there's no alias pattern in the codebase today (it's new infrastructure for one command), and I'd rather see `thv vmcp serve` ship and prove itself for a release before deciding whether `thv group serve` pulls its weight. If users reach for it naturally after `thv group ls`, we add it.

**5. Phase 4 scope signal.** Legitimate concern on paper, but Phase 4 is smaller than it reads. Most of it is wiring up code that already exists in the toolhive repo: `pkg/vmcp/optimizer/` has the FTS5 store and session-scoped indexing, `pkg/vmcp/cli/embedding_manager.go` has the full TEI container lifecycle, and `cmd/thv/app/vmcp.go` already wires the `--optimizer*` and `--embedding-*` flags. What's genuinely new is the pre-warming, `thv vmcp cleanup` with the startup orphan scan, owner-only cleanup, and the first-run UX phases. Added an "Implementation maturity note" at the top of Phase 4 so reviewers treat it as integration and gap-filling rather than a greenfield feature. If it still feels structurally heavy after that, I'm happy to spin the pre-warming + cleanup pieces off as a follow-up RFC, but I'd rather keep the story cohesive in one place.

Will push the RFC changes in the next commit. Thanks again for the thorough review.

PR #60, opened 2026-03-25
RFC-0060: Starlark Session Initialization for vMCP

jerm-dro:
Introduce a Starlark-based session initialization script for vMCP. A single script runs once per session, receives discovered backends and their capabilities, and calls `publish()` to declare what the agent sees — optionally wrapping handlers with additional logic. Existing config knobs remain fully supported, but customization of vMCP behavior can now be exactly tailored to the use case without adding more knobs. Increasing configurability no longer means decreasing maintainability.


rfcs/THV-0059-starlark-programmable-middleware.md
```diff
@@ -0,0 +1,868 @@
...
+vMCP's feature set is growing. Each feature has arrived with its own configuration surface:
+
+| Feature | Config surface | Introduced in |
+|---------|---------------|---------------|
+| Tool advertising filter | `aggregation.tools[].filter`, `excludeAll` | THV-0008 |
+| Tool renaming / overrides | `aggregation.tools[].overrides` | THV-0008 |
+| Conflict resolution | `aggregation.conflictResolution` | THV-0008 |
+| Composite tools | `compositeTools[]`, `compositeToolRefs[]` | THV-0008 |
+| Optimizer | `optimizer` (embedding service URL, thresholds, max results) | THV-0022 |
+| Starlark scripted tools | `scriptedTools[]`, `scriptedToolRefs[]` | THV-0051 |
```
jerm-dro: nit: this isn't shipped, but the existing composite tools config still does.

rfcs/THV-0059-starlark-programmable-middleware.md
```diff
@@ -0,0 +1,868 @@
...
+    rateLimiting:
+      perUser:
+        requestsPerWindow: 100
+        windowSeconds: 60
+
+# Option 3: Custom script (power users)
+middleware:
+  script: |
+    for meta, fn in tools():
+        publish(meta, fn)
```
jerm-dro: This doesn't make sense. Mixing and matching the legacy config with presets is confusing. It should be one or the other. There's no support for partial use of legacy config and some use of presets.

yrobla (changes requested):
Strong RFC overall — the problem framing, prior art, and alternatives section are all excellent. The `backends()` + `publish()` model is clean, and the default preset sketch makes backward compatibility concrete rather than aspirational.

Filing 3 blockers, 3 should-address items, 2 questions, and 2 nits.

rfcs/THV-0060-starlark-programmable-middleware.md
```diff
@@ -0,0 +1,995 @@
...
+    Agent->>Authz: tools/call "find_tool" {query: "github"}
+    Authz->>Authz: Check authorization
+    Authz->>Session: Invoke published handler
+    Session->>Backend: Forward to backend
+    Backend-->>Session: Result
+    Session-->>Authz: CallToolResult
+    Authz-->>Agent: CallToolResult
+```
+
+The script runs **once** per session, not per request. `publish()` calls build up the tool set. The resulting `(metadata, handler)` pairs are used to construct the `MultiSession`, which handles all subsequent `tools/list` and `tools/call` requests. The authz middleware sits between the agent and the session, filtering and gating requests at runtime.
```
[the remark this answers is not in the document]
jerm-dro: Good question. The session initialization script runs once per session — either when a session is created or when it's restored from Redis. The Starlark state is in-memory and not serialized, but can be recreated by re-running the script. Even though the script output could technically be shared across sessions today, running it once per session is a deliberate choice — it enables safe adoption of user-centric built-ins like `current_user()` in the future without requiring an architectural change. Added this clarification to the RFC.

JAORMX (commented):
I like the proposal and think that overall this is a good step to provide more customizability and programmability to vMCP. I'd be wary of fully relying on this mechanism for vMCP's programmability, API Gateways tend to offer programming languages as pluggability options (as you mentioned, Envoy's usage of Lua) but they often don't recommend them for critical paths and instead opt for other type of programmability or extensions. e.g. `ext_proc` via a sidecar which is quite standard to see in the wild or even c++ extensions since they support that too. For us it's a little tricky to know without previous data. So I'd say let's start getting it by adding this addition, but still opt for built-in middleware for extensibility since that's the paved path we have today. e.g. rate limiting should still be implemented via dedicated middleware IMO.

yrobla (commented):
Good RFC overall — the core argument (config knob interactions grow quadratically, Starlark + built-ins inverts this) is solid and the prior art section grounds it well. Comments below on a few specific areas.

PR #71, opened 2026-04-23
RFC: Client ID Metadata Document (CIMD) support

amirejaz:
## Summary

Proposes CIMD (Client ID Metadata Document) support across two phases:

- **Phase 1** — `thv run` acts as a CIMD-aware OAuth client. When a remote authorization server advertises `client_id_metadata_document_supported: true`, `thv run` presents its hosted metadata URL (`https://toolhive.dev/oauth/client-metadata.json`) as the `client_id` instead of performing a DCR round-trip.
- **Phase 2** — The embedded authorization server accepts CIMD. MCP clients (VS Code, Claude Code) can present an HTTPS URL as `client_id` and complete the full authorization flow without calling `/oauth/register`.

## Motivation

The MCP specification (2025-11-25) defines CIMD as the preferred client registration mechanism: `pre-registered credentials → CIMD → DCR → user prompt`. ToolHive currently jumps straight to DCR. This is both a spec-alignment issue and a friction-reduction issue — DCR requires a registration round-trip that CIMD eliminates entirely.

## Related issues

- stacklok/toolhive#2728 — umbrella issue
- stacklok/toolhive#4826 — proxy CIMD client (Phase 1)
- stacklok/toolhive#4825 — embedded auth server (Phase 2)

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rfcs/THV-0071-cimd-support.md
+++ b/rfcs/THV-0071-cimd-support.md
@@ -3,7 +3,7 @@
 - **Status**: Draft
 - **Author(s)**: Muhammad Amir Ejaz (@amirejaz)
 - **Created**: 2026-04-22
-- **Last Updated**: 2026-04-23
+- **Last Updated**: 2026-04-27
 - **Target Repository**: toolhive
 - **Related Issues**: [toolhive#2728](https://github.com/stacklok/toolhive/issues/2728), [toolhive#4825](https://github.com/stacklok/toolhive/issues/4825), [toolhive#4826](https://github.com/stacklok/toolhive/issues/4826)
 
@@ -88,24 +88,17 @@ type AuthServerInfo struct {
 }
 ```
 
-`PerformOAuthFlow` is updated to follow the new registration priority. All CIMD flows must use PKCE with `code_challenge_method=S256` — PKCE is mandatory per MCP 2025-11-25 and is orthogonal to the client registration method:
+`PerformOAuthFlow` is updated to follow the new registration priority. All CIMD flows must use PKCE with `code_challenge_method=S256` — PKCE is mandatory per MCP 2025-11-25 and is orthogonal to the client registration method.
+
+DCR is gated today by `shouldDynamicallyRegisterClient`, which checks `config.ClientID == "" && config.ClientSecret == ""`. The CIMD branch sets `config.ClientID` before that gate, so DCR does not fire when CIMD is used:
 
 ```go
-switch {
-case hasCachedCredentials(config):
-    // reuse stored client_id and refresh token — no registration needed
-
-case info.ClientIDMetadataDocumentSupported:
-    // CIMD: use our metadata URL as client_id directly; no registration call needed
+if !hasCachedCredentials(config) && info.ClientIDMetadataDocumentSupported {
+    // Set client_id before the DCR gate so shouldDynamicallyRegisterClient
+    // does not fire — PKCE (S256) is required regardless of registration method
     config.ClientID = cimd.ToolHiveClientMetadataDocumentURL
-    // PKCE (S256) is required regardless of registration method
-
-default:
-    // DCR fallback
-    if err := handleDynamicRegistration(ctx, issuer, config); err != nil {
-        return nil, err
-    }
 }
+// existing DCR gate and flow follow unchanged
 ```
 
 `client_id_metadata_document_supported` is parsed from the AS discovery document regardless of which discovery format was used to locate the AS (RFC 8414 or OIDC `.well-known/openid-configuration`).
@@ -114,7 +107,7 @@ default:
 
 CIMD carries no credentials — the `client_id` is the URL itself and nothing expires or rotates. A rejection from the AS means the server either does not support CIMD despite advertising it, or could not fetch the metadata document. It does not indicate expired or invalid credentials.
 
-If the AS rejects the CIMD `client_id` at the authorization request stage, the handler retries using DCR. The retry triggers on any of the following conditions: an `invalid_client` or `unauthorized_client` OAuth error code, an `invalid_request` error, or a raw HTTP 400/401 with no parseable error body. The fallback does not trigger at the token exchange stage — if authorization succeeds but token exchange fails, that is a separate error condition.
+If the AS rejects the CIMD `client_id` at the authorization request stage, the handler retries using DCR. The retry triggers on `invalid_client` or `unauthorized_client` OAuth error codes, or a raw HTTP 400/401 with no parseable RFC 6749 error body. `invalid_request` is not a trigger — that error indicates a malformed request, not a client identity rejection, and must surface as a real error rather than silently downgrading to DCR. The fallback does not trigger at the token exchange stage — if authorization succeeds but token exchange fails, that is a separate error condition.
 
 **`pkg/auth/remote/config.go`** (modified):
 
@@ -151,7 +144,7 @@ The document above must be publicly reachable at `https://toolhive.dev/oauth/cli
 - Valid TLS certificate — self-signed is not acceptable; third-party authorization servers will verify it
 - HTTP 200 OK on success — any other status code is treated as an error by the fetching server
 - Response body must be valid JSON containing at least `client_id` and `redirect_uris`
-- `client_id` in the document must exactly equal the serving URL — strict string comparison, no normalization
+- `client_id` in the document must exactly equal the serving URL — strict string comparison as mandated by the CIMD spec. While DNS hostnames are case-insensitive, normalising before comparison would require full IDN A-label conversion to avoid homograph bypasses, adding complexity for a case that well-implemented clients never produce. Document producers are expected to ensure their `client_id` exactly matches the URL they serve from.
 - URL must contain a path component (e.g. `/oauth/client-metadata.json`), not a bare domain
 - `Content-Type: application/json` — `application/<custom>+json` variants are also acceptable per spec, but `application/json` is the conventional choice
 - High availability — this URL is on the authorization hot path; downtime blocks `thv run` from authenticating against any CIMD-supporting server
@@ -250,9 +243,8 @@ type ClientMetadataDocument struct {
 func FetchClientMetadataDocument(ctx context.Context, rawURL string) (*ClientMetadataDocument, error)
 
 // ValidateClientMetadataDocument validates required fields, redirect_uri schemes,
-// and enforces self-referential binding: doc.ClientID must equal fetchedFrom after
-// RFC 3986 canonical normalization (lowercase scheme/host, decode unreserved chars,
-// strip default ports).
+// and enforces self-referential binding: doc.ClientID must exactly equal fetchedFrom
+// via strict string comparison per the CIMD spec.
 func ValidateClientMetadataDocument(doc *ClientMetadataDocument, fetchedFrom string) error
 ```
 
@@ -268,47 +260,51 @@ CIMD fetches also respect the CA certificate configured via `thv config set-ca-c
 
 **`pkg/authserver/storage/cimd_decorator.go`** (new file):
 
-A storage decorator wraps the existing `ClientRegistry` (fosite's `ClientManager` interface) to transparently intercept `GetClient` calls. It checks the inner storage first and fetches a CIMD document only on a miss:
+`authserver.New()` takes `storage.Storage` — a ~15-method interface. The decorator embeds the full base storage and only overrides `GetClient`, delegating all other methods transparently. It uses a `singleflight.Group` to collapse concurrent fetches for the same URL into one outbound request, preventing thundering-herd on the anonymous `/authorize` endpoint where fosite calls `GetClient` before validating anything else.
 
 ```go
-// CIMDStorageDecorator wraps ClientRegistry and handles CIMD client_id values.
-// Results are cached in an LRU cache with TTL derived from the Cache-Control
-// response header, falling back to a configurable default.
+// CIMDStorageDecorator embeds storage.Storage and overrides GetClient only.
+// All other Storage methods delegate to base unchanged.
 type CIMDStorageDecorator struct {
-    base    ClientRegistry
-    cache   *lru.Cache[string, cacheEntry] // LRU-bounded; TTL per entry from Cache-Control
-    maxSize int                            // default: 256 entries
-    ttl     time.Duration                  // fallback TTL when Cache-Control is absent
+    storage.Storage                             // embed full interface
+    sf      singleflight.Group                 // one fetch per URL under concurrent load
+    cache   *lru.Cache[string, cacheEntry]     // LRU-bounded; TTL per entry from Cache-Control
+    maxSize int                                // default: 256 entries
+    ttl     time.Duration                      // fallback TTL when Cache-Control is absent
     enabled bool
     mu      sync.RWMutex
 }
 
 func (d *CIMDStorageDecorator) GetClient(ctx context.Context, id string) (fosite.Client, error) {
     if !d.enabled || !cimd.IsClientIDMetadataDocumentURL(id) {
-        return d.base.GetClient(ctx, id)
+        return d.Storage.GetClient(ctx, id)
     }
-    return d.fetchOrCached(ctx, id)
+    return d.fetchOrCached(ctx, id) // uses sf.Do to deduplicate concurrent fetches
 }
 
-// RegisterClient always delegates to base — DCR clients continue to work.
-func (d *CIMDStorageDecorator) RegisterClient(ctx context.Context, client fosite.Client) error {
-    return d.base.RegisterClient(ctx, client)
+// Unwrap returns the underlying storage so type assertions (e.g. legacy
+// RedisStorage migration in server_impl.go) can reach the base type.
+func (d *CIMDStorageDecorator) Unwrap() storage.Storage {
+    return d.Storage
 }
 ```
+
+**Wiring note:** `pkg/authserver/server_impl.go:146` contains a `stor.(*storage.RedisStorage)` type assertion for a legacy data migration. The migration must run against the unwrapped storage before the decorator is applied, or the assertion site must be updated to call `Unwrap()` first.
+
+**Call site coverage:** `GetClient` is called in two places in `pkg/authserver/server/handlers/callback.go` (around lines 168 and 259) in addition to the fosite interception. Both call sites use `h.storage`, so wiring the decorator at the `authserver.New()` level covers both automatically.
 
 The decorator constructs a `fosite.DefaultClient` from the fetched `ClientMetadataDocument`. For documents containing `http://localhost` redirect URIs, it wraps the result in the existing `LoopbackClient` (`pkg/authserver/server/registration/client.go`) to enable port-agnostic loopback matching per RFC 8252 §7.3.
 
-**`pkg/authserver/server/handlers/discovery.go`** (modified):
+**`pkg/oauth/discovery.go`** (modified):
 
-Both well-known endpoints advertise the new capability when `CIMDEnabled: true`:
+`AuthorizationServerMetadata` lives in the shared `pkg/oauth` package. The field is added there; the handler in `pkg/authserver/server/handlers/discovery.go` only populates it:
 
 ```go
-// Added to AuthorizationServerMetadata (served at both
-// /.well-known/oauth-authorization-server and /.well-known/openid-configuration)
+// Added to AuthorizationServerMetadata in pkg/oauth/discovery.go
 ClientIDMetadataDocumentSupported bool `json:"client_id_metadata_document_supported,omitempty"`
 ```
 
-Advertising in both endpoints ensures clients using either RFC 8414 or OIDC discovery can detect CIMD support and skip the DCR step automatically.
+Advertising in both well-known endpoints (`/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`) ensures clients using either RFC 8414 or OIDC discovery can detect CIMD support and skip the DCR step automatically.
 
 **Embedded auth server configuration** (new opt-in field, default disabled):
 
@@ -358,7 +354,7 @@ Threats marked **Phase 2** apply only when the embedded authorization server fet
 
 | Threat | Phase | Description | Mitigation |
 |--------|-------|-------------|------------|
-| SSRF via CIMD URL | Phase 2 | Attacker presents a CIMD `client_id` that resolves to an internal service | DNS-after-resolve check blocks RFC 1918, loopback, and link-local addresses |
+| SSRF via CIMD URL | Phase 2 | Attacker presents a CIMD `client_id` that resolves to an internal service | DNS-after-resolve check using `net.IP.IsGlobalUnicast()` allowlist; blocks RFC 1918, loopback, link-local, shared address space (100.64.0.0/10), IPv4-mapped IPv6, and documentation ranges |
 | DoS via large response | Phase 2 | Attacker's metadata server returns an oversized response body | 10 KB body cap via `io.LimitReader` before JSON parsing |
 | DoS via slow metadata server | Phase 2 | Attacker serves metadata very slowly to tie up connections | 5-second HTTP timeout on all CIMD fetch requests |
 | Metadata spoofing | Phase 2 | Attacker serves a document claiming to be a different client | `client_id` in the document must equal the URL it was fetched from (self-referential binding) |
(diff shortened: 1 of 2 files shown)


rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+
+`AuthServerInfo` gains a new field populated when the remote AS discovery document is parsed:
+
+```go
+type AuthServerInfo struct {
+    Issuer                            string
+    AuthorizationURL                  string
+    TokenURL                          string
+    RegistrationEndpoint              string
+    ClientIDMetadataDocumentSupported bool // parsed from AS metadata
```
jhrozek, line 87: A bit of a tangent, but have we thought about consolidating the protocol-level modules into pkg/oauth instead? Longer-term we should move pkg/oauth to toolhive-core and be reusable by other repos so that we prevent code duplication across projects.

I don't know if this particular change is possible to do, but my gut feeling is that "generic CIMD support" should probably be part of pkg/oauth

rfcs/THV-XXXX-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+    // DCR fallback
+    if err := handleDynamicRegistration(ctx, issuer, config); err != nil {
+        return nil, err
+    }
+}
+```
+
+**`pkg/auth/remote/handler.go`** (modified):
+
+If the AS rejects the CIMD `client_id` with an `invalid_client` error, the handler retries the authorization flow using DCR — graceful fallback per issue #4826 acceptance criteria.
```
jhrozek: honest question, are cimd creds expiring or rotated? Is there a way to distinguish between "CIMD is not supported" and "CIMD creds no longer valid"?

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+
+If the AS rejects the CIMD `client_id` with an `invalid_client` error, the handler retries the authorization flow using DCR — graceful fallback per issue #4826 acceptance criteria.
+
+**`pkg/auth/remote/config.go`** (modified):
+
+```go
+CachedCIMDClientID string // set when CIMD was used; avoids re-detection on restart
+```
+
+**`toolhive-client-metadata.json`** (new static file, checked into repo):
```
jhrozek, line 118: not a hard requirement for now, we havent' had this ask but I wonder if it would be possible to configure a custom client metadata file location? Thinking about disconnected/offline environments that cant' fetch from the internet but only can fetch from corp network

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+`application_type: "native"` correctly classifies ToolHive as a native CLI application. Two redirect URIs are included: `http://localhost/callback` for port-agnostic loopback matching (RFC 8252 §7.3) and `http://localhost:8855/callback` as a fixed-port fallback. `thv run` binds to port 8855 by default and falls back to an OS-assigned port if unavailable.
+
+#### Hosting the Client Metadata Document
+
+The document above must be publicly reachable at `https://toolhive.dev/oauth/client-metadata.json` **before Phase 1 ships**. This is a hard infrastructure prerequisite — `thv run` cannot use CIMD until the URL is live.
+
+**Serving requirements:**
+
+- `https` scheme only — HTTP is not permitted
+- Valid TLS certificate — self-signed is not acceptable; third-party authorization servers will verify it
```
jhrozek, line 144: do we have a place in the code to inject custom CA certs?

rfcs/THV-XXXX-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+- `https` scheme only — HTTP is not permitted
+- Valid TLS certificate — self-signed is not acceptable; third-party authorization servers will verify it
+- HTTP 200 OK on success — any other status code is treated as an error by the fetching server
+- Response body must be valid JSON containing at least `client_id` and `redirect_uris`
+- `client_id` in the document must exactly equal the serving URL — strict string comparison, no normalization
+- URL must contain a path component (e.g. `/oauth/client-metadata.json`), not a bare domain
+- `Content-Type: application/json` — `application/<custom>+json` variants are also acceptable per spec, but `application/json` is the conventional choice
+- High availability — this URL is on the authorization hot path; downtime blocks `thv run` from authenticating against any CIMD-supporting server
+
+**Note on response size:** The IETF draft recommends document producers keep responses under 5 KB. As a consumer, ToolHive's fetch code enforces a 10 KB hard cap as a defence-in-depth measure to reject maliciously large responses — these are distinct limits with different purposes.
```
jhrozek: I vaguely remember that CIMD can support logos, are they a link or inline? If inline 10kB might be too little

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+
+Phase 2 adds the document type, fetch, and validation logic — only needed when ToolHive acts as the authorization server that receives CIMD `client_id` values from connecting clients.
+
+```go
+// ClientMetadataDocument represents a Client ID Metadata Document per
+// draft-ietf-oauth-client-id-metadata-document.
+//
+// Required: ClientID, RedirectURIs.
+// Recommended: ClientName, LogoURI, ClientURI.
+type ClientMetadataDocument struct {
```
jhrozek, line 219: this sounds like something belonging to pkg/oauth

rfcs/THV-XXXX-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+### Mitigations
+
+The SSRF protection logic is extracted into `pkg/auth/oauth/ssrf.go` (introduced in Phase 2) as a standalone utility so it can be reused by any future HTTP-fetching code in the auth layer without duplicating the IP range checks.
+
+## Compatibility
+
+### Backward Compatibility
+
+- DCR remains fully functional as a fallback for both the embedded server and the proxy. No existing configuration or deployments require changes.
+- The embedded server: CIMD is opt-in (`enabled: false` by default). Existing DCR clients connecting to the embedded server are completely unaffected.
```
jhrozek: do we need an opt-in? WHy not an opt-out?

rfcs/THV-XXXX-cimd-support.md
```diff
@@ -0,0 +1,474 @@
...
+- Modified: `pkg/auth/remote/handler.go` — graceful fallback to DCR when AS rejects CIMD `client_id`
+
+### Phase 2: Embedded auth server CIMD support — closes #4825
+
+Begins after Phase 1 ships. Adds the server-side counterpart so MCP clients (VS Code, Claude Code) connecting to ToolHive-managed servers can use an HTTPS URL as `client_id` and complete the full auth flow without a prior DCR call.
+
+This phase introduces the shared CIMD fetch and validation logic (not needed by Phase 1) alongside the embedded server integration:
+
+- New: `pkg/auth/oauth/cimd.go` — extend with `FetchClientMetadataDocument`, `ValidateClientMetadataDocument`
+- New: `pkg/auth/oauth/ssrf.go` — SSRF protection utility shared across auth layer
```
jhrozek: can't we reuse the existing SSRF protections in the round-tripper?

rfcs/THV-0071-cimd-support.md
amirejaz: Good point — updated. cimd.go is now placed in pkg/oauth alongside the other protocol utilities. The long-term toolhive-core consolidation is a separate initiative.

rfcs/THV-XXXX-cimd-support.md
amirejaz: CIMD carries no credentials — the client_id is the URL itself and nothing expires or rotates. A rejection from the AS means the server either does not support CIMD despite advertising it, or could not fetch the metadata document. The RFC now clarifies this, and the fallback trigger covers the full set of error conditions: invalid_client, unauthorized_client, invalid_request, or a raw HTTP 400/401.

rfcs/THV-0071-cimd-support.md
amirejaz: Acknowledged as a follow-up enhancement. The RFC now explicitly notes that a per-deployment URL override (for air-gapped or corporate environments) is out of scope for this RFC.

rfcs/THV-0071-cimd-support.md
amirejaz: Yes — pkg/networking/http_client.go already has HttpClientBuilder with CA bundle loading built in. The CIMD fetch transport (Phase 2) uses that same builder, so the CA cert configured via thv config set-ca-cert is respected automatically. The RFC has been updated to reflect this.

rfcs/THV-XXXX-cimd-support.md
amirejaz: logo_uri is a URL reference per RFC 7591 (from which CIMD inherits its metadata fields), not inline image data. The 10 KB cap applies to the JSON metadata document itself, not any linked resources. The RFC now makes this explicit.

rfcs/THV-0071-cimd-support.md
amirejaz: Updated — FetchClientMetadataDocument and the full ClientMetadataDocument type now live in pkg/oauth/cimd.go alongside the URL constant and IsClientIDMetadataDocumentURL helper.

rfcs/THV-XXXX-cimd-support.md
amirejaz: The opt-in applies to Phase 2 (embedded AS) only — Phase 1 (thv run) requires no configuration flag and activates CIMD automatically when the remote AS advertises support. The reason Phase 2 is opt-in: it introduces outbound HTTPS fetching from the authorization server, which is new network behaviour not present in any existing deployment. The RFC now states this explicitly.

rfcs/THV-XXXX-cimd-support.md
amirejaz: Yes — pkg/networking/http_client.go already has exactly this: HttpClientBuilder with a DialContext-based SSRF check (protectedDialerControl) and a ValidatingTransport round-tripper. Phase 2 reuses that infrastructure rather than introducing new code. pkg/networking will be extended with the missing ranges (100.64.0.0/10, IPv4-mapped IPv6, documentation ranges). The RFC no longer proposes a new ssrf.go file.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+| Compromised hosting | Phase 1 & 2 | Attacker gains control of the HTTPS origin serving a CIMD document (CDN, GitHub Pages, object storage) and replaces the metadata | **Accepted residual risk** — CIMD client trust is equivalent to controlling the HTTPS URL. TLS ensures transport integrity but does not prevent the legitimate host from being compromised. This is a fundamental property of the CIMD trust model. |
+
+### Authentication and Authorization
+
+CIMD documents are intentionally public — they contain no credentials. Trust is established through:
+
+1. **TLS certificate validation** (Phase 2): All CIMD fetches use standard Go TLS verification; self-signed certificates are rejected.
+2. **Self-referential binding** (Phase 2): The `client_id` field inside the fetched document must equal the URL from which it was fetched after RFC 3986 normalization (lowercase scheme and host, decode unreserved percent-encoded characters, strip default ports). A document hosted at `https://example.com/client.json` claiming `"client_id": "https://attacker.com/client.json"` is rejected.
```
jhrozek: Why RFC 3986 canonical normalization here, rather than simple string comparison?

The CIMD draft mandates simple string comparison (the `client_id` MUST match the URL using simple string comparison). The RFC's own producer-side requirement at L154 says "strict string comparison, no normalization" — so the producer rule and the consumer rule (here and at L394) currently contradict each other.

Was this a deliberate tightening for security? Asking because:

1. An AS strictly following the draft will reject documents this validator accepts (interop concern).
2. If we keep normalization, the current ruleset is incomplete — missing IDN→A-label conversion (`golang.org/x/net/idna`), path dot-segment collapse, fragment/userinfo/query handling. Without IDN handling, `https://аpple.com/x` (Cyrillic "а") can let the fetch path and the binding check resolve differently — a self-binding bypass.

Either path works; just want to make sure it's an explicit decision and consistently applied between L154 and L378/L394.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+| DNS rebinding | Phase 2 | Domain resolves to public IP for the SSRF check, then re-resolves to private IP for the actual connection | SSRF check performed on the IP used for the actual connection via custom `http.Transport` |
+| SSRF via redirect chain | Phase 2 | Metadata server issues redirects to an internal target | One-hop redirect limit; redirect destination subject to the same SSRF IP validation |
+| Scope escalation | Phase 2 | CIMD document claims broader scopes than server policy allows | Scopes intersected with server-configured allowed scopes at token issuance |
+| Client impersonation | Phase 1 & 2 | Attacker hosts a CIMD document mimicking a legitimate client | `client_id` self-referential binding + TLS certificate validation |
+| LRU cache flooding | Phase 2 | Attacker submits many distinct CIMD URLs to fill the 256-entry LRU, evicting legitimate entries and triggering repeated outbound fetches | LRU eviction bounds memory; consider per-source-IP rate limiting at the authorization endpoint as a follow-up hardening measure |
```
jhrozek, line 366: The LRU flooding row says rate limiting is a follow-up, but I'm a little worried about that. `/authorize` is anonymous (no middleware on it today) and fosite calls `GetClient` before it validates anything else, so any malformed request with an https-shaped client_id makes us fetch. With a 256-entry LRU someone can evict everything by submitting 256 different URLs, and there's no single-flight so a parallel burst with the same URL fans out into N fetches.

Could we put singleflight in the decorator before Phase 2 ships? It's a few lines and kills the thundering-herd case. A small concurrent-fetch semaphore as a backstop would be nice too. Both fit inside the decorator without touching the router.

Rate limiting and negative caching as a follow-up seems fine to me.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+
+`pkg/networking` will be extended with the following missing IP ranges: shared address space (100.64.0.0/10 per RFC 6598), IPv4-mapped IPv6 addresses (::ffff:0:0/96), and documentation ranges (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24). The implementation uses `net.IP.IsGlobalUnicast()` as the primary allowlist — only globally routable IPs are permitted — covering current and future special-use ranges.
+
+The CIMD fetch transport must be constructed with `DisableKeepAlives: true` to prevent pooled connections from bypassing the SSRF check, and the `DialContext` hook validates the IP that Go's Happy Eyeballs actually dials (not a separately-resolved address). A one-hop redirect limit applies; redirect destinations are subject to the same IP validation.
+
+CIMD fetches also respect the CA certificate configured via `thv config set-ca-cert` — the `HttpClientBuilder` CA bundle loading is used, appended to the system cert pool.
+
+**`pkg/authserver/storage/cimd_decorator.go`** (new file):
+
```
jhrozek, line 262: Small thing on the decorator shape — `authserver.New()` takes `storage.Storage` (the full interface, not just `ClientRegistry`), so the decorator either needs to embed the base storage and only override `GetClient`, or we need to wire it into fosite's provider separately from the rest of the storage. Worth showing the embedding in the snippet so it's clear this is a ~15-method wrap, not a 2-method one.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+    AS->>M: GET https://vscode.dev/oauth/client-metadata.json
+    M-->>AS: {client_id, redirect_uris, ...}
+    AS->>AS: validate + cache → build fosite.Client
+    AS-->>C: redirect to upstream IDP
+    Note over C,AS: ... user authenticates ...
+    C->>AS: POST /oauth/token
+    AS-->>C: {access_token: JWT}
+```
+
+The interception happens at `GetClient` — the point fosite calls into storage to validate the client during the authorize request. The `CIMDStorageDecorator` wraps this call: if the `client_id` is an HTTPS URL and CIMD is enabled, it fetches the document instead of looking up a registered UUID.
```
jhrozek, line 207: Heads-up for the implementation — there's a second `GetClient` site in `pkg/authserver/server/handlers/callback.go` (around lines 168 and 259, look for `h.storage.GetClient`). The flow described here intercepts at fosite, but the callback handler calls into storage directly so it'd bypass the decorator unless we wire it in there too. Probably worth a sentence noting both call sites need to see the wrapped storage.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+}
+
+// RegisterClient always delegates to base — DCR clients continue to work.
+func (d *CIMDStorageDecorator) RegisterClient(ctx context.Context, client fosite.Client) error {
+    return d.base.RegisterClient(ctx, client)
+}
+```
+
+The decorator constructs a `fosite.DefaultClient` from the fetched `ClientMetadataDocument`. For documents containing `http://localhost` redirect URIs, it wraps the result in the existing `LoopbackClient` (`pkg/authserver/server/registration/client.go`) to enable port-agnostic loopback matching per RFC 8252 §7.3.
+
```
jhrozek, line 297: One more landmine for the implementer — `pkg/authserver/server_impl.go:146` does a `stor.(*storage.RedisStorage)` type assertion to run a legacy data migration. If we wrap with the decorator first, that assertion fails silently and the migration no-ops. Either run the migration before wrapping, or have the decorator expose `Unwrap()` and update the assertion site. Probably worth calling out so it doesn't slip through.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+```
+
+Advertising in both endpoints ensures clients using either RFC 8414 or OIDC discovery can detect CIMD support and skip the DCR step automatically.
+
+**Embedded auth server configuration** (new opt-in field, default disabled):
+
+```yaml
+authServer:
+  cimd:
+    enabled: false        # opt-in; set to true to accept HTTPS URLs as client_id
```
jhrozek, line 314: The implementation-plan bullet that says "Modified: embedded auth server config — add CIMDEnabled flag" is doing a lot of work — the YAML actually has to thread through `authserver.RunConfig` → `Config` → `AuthorizationServerParams` → `Handler` (so `buildOAuthMetadata` can set the discovery field) → the decorator constructor. None of those structs have a CIMD field today. Worth listing the touched files in the implementation plan so we don't underestimate.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+    TokenURL                          string
+    RegistrationEndpoint              string
+    ClientIDMetadataDocumentSupported bool // parsed from AS metadata
+}
+```
+
+`PerformOAuthFlow` is updated to follow the new registration priority. All CIMD flows must use PKCE with `code_challenge_method=S256` — PKCE is mandatory per MCP 2025-11-25 and is orthogonal to the client registration method:
+
+```go
+switch {
```
jhrozek: The switch sketch here doesn't quite map onto how `PerformOAuthFlow` is structured today — DCR is gated by `shouldDynamicallyRegisterClient` checking `ClientID == "" && ClientSecret == ""` (around L530), and the actual call happens further down. To make the CIMD branch work cleanly we'd either set `config.ClientID = ToolHiveClientMetadataDocumentURL` before that gate so DCR doesn't fire, or refactor the function. Worth showing which.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+### Phase 2: Embedded auth server CIMD support — closes #4825
+
+Begins after Phase 1 ships. Adds the server-side counterpart so MCP clients (VS Code, Claude Code) connecting to ToolHive-managed servers can use an HTTPS URL as `client_id` and complete the full auth flow without a prior DCR call.
+
+This phase introduces the shared CIMD fetch and validation logic (not needed by Phase 1) alongside the embedded server integration:
+
+- New: `pkg/oauth/cimd.go` — extend with `FetchClientMetadataDocument`, `ValidateClientMetadataDocument`
+- Modified: `pkg/networking/http_client.go` — extend existing `HttpClientBuilder` with missing IP ranges (100.64.0.0/10, IPv4-mapped IPv6, documentation ranges); used by CIMD fetch transport
+- New: `pkg/authserver/storage/cimd_decorator.go` — storage decorator wrapping `GetClient`
+- Modified: `pkg/authserver/server/handlers/discovery.go` — advertise `client_id_metadata_document_supported` in both well-known endpoints
```
jhrozek: Tiny correction on the implementation-plan bullet — the `AuthorizationServerMetadata` struct itself lives in `pkg/oauth/discovery.go` (shared package), so the field add goes there; `pkg/authserver/server/handlers/discovery.go` only populates it.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+}
+```
+
+`client_id_metadata_document_supported` is parsed from the AS discovery document regardless of which discovery format was used to locate the AS (RFC 8414 or OIDC `.well-known/openid-configuration`).
+
+**`pkg/auth/remote/handler.go`** (modified):
+
+CIMD carries no credentials — the `client_id` is the URL itself and nothing expires or rotates. A rejection from the AS means the server either does not support CIMD despite advertising it, or could not fetch the metadata document. It does not indicate expired or invalid credentials.
+
+If the AS rejects the CIMD `client_id` at the authorization request stage, the handler retries using DCR. The retry triggers on any of the following conditions: an `invalid_client` or `unauthorized_client` OAuth error code, an `invalid_request` error, or a raw HTTP 400/401 with no parseable error body. The fallback does not trigger at the token exchange stage — if authorization succeeds but token exchange fails, that is a separate error condition.
```
jhrozek: Coming back to this — I get the rationale for being permissive, but I'm a little uneasy about `invalid_request` specifically. That's an error about request shape, not client identity. An AS that doesn't know about CIMD will return `invalid_client` per RFC 6749 §5.2; `invalid_request` would more typically mean we sent a malformed request, which we'd then silently mask by downgrading to DCR. A misbehaving AS could also use it to force a downgrade. Could we narrow to `invalid_client` / `unauthorized_client` only and let `invalid_request` bubble up as a real error?

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+func ValidateClientMetadataDocument(doc *ClientMetadataDocument, fetchedFrom string) error
+```
+
+**`pkg/networking` (enhanced, existing package):**
+
+CIMD fetches reuse the existing `HttpClientBuilder` in `pkg/networking/http_client.go`, which already provides a `DialContext`-based SSRF check via `protectedDialerControl` and a `ValidatingTransport` round-tripper. No new SSRF utility file is introduced.
+
+`pkg/networking` will be extended with the following missing IP ranges: shared address space (100.64.0.0/10 per RFC 6598), IPv4-mapped IPv6 addresses (::ffff:0:0/96), and documentation ranges (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24). The implementation uses `net.IP.IsGlobalUnicast()` as the primary allowlist — only globally routable IPs are permitted — covering current and future special-use ranges.
+
+The CIMD fetch transport must be constructed with `DisableKeepAlives: true` to prevent pooled connections from bypassing the SSRF check, and the `DialContext` hook validates the IP that Go's Happy Eyeballs actually dials (not a separately-resolved address). A one-hop redirect limit applies; redirect destinations are subject to the same IP validation.
```
jhrozek, line 257: Quick implementation note — the existing `HttpClientBuilder` doesn't set `DisableKeepAlives` and doesn't expose a setter for it, so the CIMD fetch can't get that property just by building from the helper. Easiest path is probably adding `WithDisableKeepAlives(bool)` to the builder; otherwise CIMD would have to construct its own transport and we'd lose some of the consolidation benefit.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+
+**Phase 1** — `pkg/auth/remote/config.go` gains:
+
+```go
+CachedCIMDClientID string // set when CIMD was used; avoids re-detection on restart
+```
+
+This follows the existing persistence pattern for `CachedClientID` (DCR client ID) and `CachedRefreshTokenRef`. It is non-sensitive (a public URL).
+
+**Refresh token behavior**: during token refresh, `client_id=https://toolhive.dev/oauth/client-metadata.json` is sent to the token endpoint as required by RFC 6749. The CIMD document is not re-fetched on refresh — the client identity from the original authorization flow is reused. If the document has changed since the access token was issued, the change takes effect only on the next full authorization flow.
```
jhrozek, line 345: Worth saying explicitly here — since the static doc has `token_endpoint_auth_method: "none"`, ToolHive is a public client, and OAuth 2.1 + MCP 2025-11-25 require the AS to rotate refresh tokens on every refresh. Whoever implements the refresh code needs to replace `CachedRefreshTokenRef` with the new token from each refresh response (and must not reuse the prior one). One-line addition to the paragraph would head off a future bug.

rfcs/THV-0071-cimd-support.md
```diff
@@ -0,0 +1,501 @@
...
+//
+// Required: ClientID, RedirectURIs.
+// Recommended: ClientName, LogoURI, ClientURI.
+type ClientMetadataDocument struct {
+    // Required
+    ClientID     string   `json:"client_id"`     // must exactly equal the URL this document is served from
+    RedirectURIs []string `json:"redirect_uris"` // exact URIs only; no wildcards permitted
+
+    // Recommended
+    ClientName string `json:"client_name,omitempty"`
```
jhrozek, line 225: MCP 2025-11-25 actually says `client_name` is required (alongside `client_id` and `redirect_uris`), not just recommended. Worth promoting it to Required in the struct comment so the validator rejects docs missing it.

jhrozek (commented):
Second-wave review based on a deeper cross-check against the toolhive code. Most of these are concerns about the Phase 2 decorator design and a few smaller nits. Skipping straight to inline comments below.

jhrozek (approved):
Good work on the revisions — thanks for the thorough turnaround.

jhrozek:
@amirejaz wanna merge this? :-)

PR #72, opened 2026-05-13
RFC: Graceful Degradation for Multi-Upstream OAuth in vMCP

tgrunnagle:
Adds an RFC proposing opt-in partial-completion semantics for the
embedded authorization server's multi-upstream OAuth chain in
VirtualMCPServer.

Related issue: stacklok/toolhive#5162

> **Note:** This PR body was updated post-merge to reflect the
> RFC's final design after review feedback. See follow-up commits
> `fde11e0` and `031eda4` on the same branch for the details.

## Why

Today the embedded auth server's multi-upstream chain is all-or-nothing:
one upstream IdP outage or one declined consent screen invalidates every
collected token and locks the user out of every backend on the vMCP —
including backends that have nothing to do with the failed upstream.
Operators aggregating backends across multiple SaaS IdPs (e.g. github +
slack + google) see the whole vMCP appear down for what should be a
single-provider problem.

## Key design decisions

- **Single per-upstream `optional` flag** (default `false`) on
  `UpstreamProviderConfig`. Marking any provider `optional: true` opts
  the deployment into partial completion for that provider; leaving
  every provider at the default reproduces today's all-or-nothing
  behavior exactly. No top-level mode toggle.
- **Primary identity provider** — already named by
  `authzConfig.inline.primaryUpstreamProvider` — is always required;
  admission webhook rejects configs that set `optional: true` on it.
- **Chain completion** walks every upstream and completes when every
  required upstream has a token; optional upstreams that error or are
  declined are recorded as session-scoped skip rows.
- **Skip rows reuse the existing `UpstreamTokens` storage row.** Two
  new scalar fields (`Skipped bool`, `SkippedReason string`) on the
  existing row; the `UpstreamTokenStorage` interface is unchanged. The
  existing `InProcessService.GetAllValidTokens` / `GetValidToken`
  filter skip rows out of the live-token map, so `identity.UpstreamTokens`
  consumers behave exactly as today when no rows are skipped.
- **No `/authorize` restart-all branch.** The existing handler already
  mints a fresh `SessionID` per call, so re-running `/authorize` is
  structurally a new session — prior session rows age out under their
  existing TTL. Per-upstream retry is explicitly rejected — identity-
  binding hazards outweigh the round-trip savings.
- **vMCP filters by skipping unauthorized backend init.** vMCP freezes
  the per-session capability set at MCP `initialize` time, so the
  filter hooks into the pre-init backend filter loop of
  `makeBaseSession`: backends whose required upstream is missing from
  `identity.UpstreamTokens` never run `initOneBackend` at all — no HTTP
  connection, no handshake, no capability listing, no entry into the
  routing table or per-session SDK tool store.
- **No new dispatcher gate.** Out-of-band revocation between
  `initialize` and call is already caught by the existing per-request
  `identity.UpstreamTokens` re-hydration plus each backend auth
  strategy's `ErrUpstreamTokenNotFound` path.
- **Recovery is whole-session.** A fresh `/authorize` produces a new
  JWT with a new `tsid`; reconnecting to vMCP with the new bearer
  triggers a fresh `initialize` and re-aggregates capabilities.
- vMCP does not extend the MCP protocol to signal filtered state.

## Out of scope

Per-upstream retry, MCP-protocol-level signaling of filtered state,
silent backend dropping on RT expiry, proactive token introspection,
dynamic upstream addition.

## Notes for reviewers

- Implementation is sequenced runtime-first → CRD-last so chain-logic
  and filtering can land and be tested before the CRD surface
  solidifies.
- All initial open questions (revocation detection, skip-row TTL,
  metric cardinality, status conditions) were resolved during drafting;
  see the body for the decisions.
- The RFC originated in the toolhive repo via Claude Code (referencing
  source line numbers); file-path references are plain text so they
  remain readable when the RFC is moved to its destination repo.

### DIFF
--- a/rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
+++ b/rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
@@ -13,10 +13,12 @@ When a VirtualMCPServer's embedded authorization server fronts multiple upstream
 Identity Providers, today's chain is all-or-nothing: any one failure invalidates
 every collected upstream token and locks the user out of every backend, even
 backends that depend only on the upstreams that already succeeded. This RFC
-introduces opt-in partial-completion semantics, controlled by a new
-`partialUpstreamAuth` knob on the embedded auth server and a `required` flag
-on each upstream provider. Backends whose required upstream is missing are
-filtered out of the vMCP's exposed catalog and refused at request time;
+introduces opt-in partial-completion semantics, controlled by a single
+`optional` flag (default `false`) on each upstream provider. Marking any
+provider `optional: true` opts the deployment into partial completion for
+that provider; leaving every provider at the default reproduces today's
+all-or-nothing behavior exactly. Backends whose required upstream is missing
+are filtered out of the vMCP's exposed catalog and refused at request time;
 re-authentication is restart-all, with full identity-consistency checks
 preserved.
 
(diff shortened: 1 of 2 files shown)


rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
+# RFC-0072: Graceful Degradation for Multi-Upstream OAuth in vMCP
+
+- **Status**: Draft
+- **Author(s)**: TBD (@github-handle)
```
jhrozek: This TBD dude seems to submit the best RFCs. We should hire him!

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+- **Related Issues**: [toolhive#5162](https://github.com/stacklok/toolhive/issues/5162)
+
+## Summary
+
+When a VirtualMCPServer's embedded authorization server fronts multiple upstream
+Identity Providers, today's chain is all-or-nothing: any one failure invalidates
+every collected upstream token and locks the user out of every backend, even
+backends that depend only on the upstreams that already succeeded. This RFC
+introduces opt-in partial-completion semantics, controlled by a new
+`partialUpstreamAuth` knob on the embedded auth server and a `required` flag
```
jhrozek: why do we have 2 flags? Wouldn't it be easier to only add the `required` flag to the individual upstreams and allow to opt out each upstream by setting the flag to false?

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+
+1. **Capability listings.** Wrap the aggregator's response to
+   `tools/list`, `resources/list`, `resources/templates/list`, and
+   `prompts/list` with a filter that drops any item whose owning backend
+   reports a non-empty `RequiredUpstreamProvider()` not present in the
+   session's `IdentityFromContext(ctx).UpstreamTokens` map
+   (pkg/vmcp/auth/strategies/upstream_inject.go:44-87).
+   Listings are computed per-request, not once at server start, so the
+   filter naturally tracks per-session state. The filter never mutates
+   the backend registry; only the per-request view is altered.
```
jhrozek: is this correct? I thought there were hooks into initialize in vMCP and not the list methods? This would mean that the client needs to fully re-initialize and re-run authorize to get the new set of resources/tools/prompts

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+that the same user binds the session — if a different user comes back, the
+existing session-binding fields
+(`UpstreamTokens.UserID`, `.UpstreamSubject`,
+pkg/authserver/storage/types.go:55-97)
+force a fresh session ID instead of letting the new identity adopt the
+old session.
+
+Session correlation: the client's `/authorize` request supplies its
+client ID and (in PKCE flows) a fresh code challenge. Session reuse is
+keyed off the existing pending-authorization correlation — no new client
```
jhrozek: how does this work exactly, does this RFC specify the mechanism? We do have the tsid, but that's not passed to the new authorize session is it?

jhrozek: Were you maybe planning to look up existing session not during authorize but during callback where we can establish the identity of the caller?

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+map[string]string` (session → provider → reason) guarded by the existing
+mutex. Redis implementation uses a hash keyed by sessionID with provider
+names as fields. The set is wiped by `DeleteUpstreamTokens` in the
+existing restart-all path.
+
+Tombstones have no independent TTL: they live for the lifetime of the
+session and are cleared only by restart-all. Retrying a previously
+declined upstream requires the client to re-run `/authorize`, which
+trips restart-all and walks the full chain again — consistent with the
+restart-all-only recovery model used elsewhere in this RFC.
```
jhrozek: why not tie them to the session of the AS-issued JWT? Won't this mean that the tombstone entries are never cleared?

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+   (default — equivalent to today) or `partialUpstreamAuth: allow`, in which
+   case per-provider `required` is honored.
+2. **Skip-and-continue on optional-upstream failure.** When an optional
+   upstream returns an RFC 6749 §4.1.2.1 error (including user-declined
+   `access_denied`), the chain records the upstream as *skipped for this
+   session* and advances to the next missing upstream instead of tearing the
+   whole flow down. The auth code is issued when every upstream has been
+   attempted and every *required* upstream has a token.
+3. **Backend filtering, fail-closed.** The vMCP request pipeline reads the
+   per-session set of available upstream tokens off the bearer-token claim
```
jhrozek: umm does it? I thought we read them from storage and then augment the Identity struct with the tokens.

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
```diff
@@ -0,0 +1,810 @@
...
+   stored token nor a skipped tombstone for the session. Chain completion
+   is now a two-part predicate evaluated after each successful leg:
+   `nextUnattemptedUpstream` returns empty *and* every required upstream
+   has a stored token. Optional upstreams are still walked so the user has
+   a chance to authorize them; only after they have been attempted (and
+   either succeeded or been recorded as skipped) does the chain consider
+   itself complete.
+
+2. **Skipped-upstream tombstone in storage.** Add a
+   `MarkUpstreamSkipped(sessionID, providerName, reason)` method to
```
jhrozek: Wouldn't it be easier to extend the `UpstreamTokens` struct with flags to mark the token (either the one being stored or retrieved) as tombstone and then just extend the existing `GetAllUpstreamTokens` and `StoreUpstreamTokens`?

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
tgrunnagle: I imagined it as a backward compatibility concern, assuming 'required' defaults to false. An operator that upgrades versions without changing the manifest would get the existing behavior. I realize now that rather than two flags, we could add an 'optional' (default false) flag to each upstream and achieve the same existing behavior. I'll update the doc.

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
tgrunnagle: I was imagining it could read the bearer token from the header, but we should just maintain the existing behavior (new tsid per `/authorize` call)

rfcs/THV-0072-vmcp-multi-upstream-partial-auth.md
tgrunnagle: Updated storage to be a flag on existing `UpstreamTokens` entries

jhrozek (approved):
very nice thank you

PR #73, opened 2026-05-21
RFC: vMCP session binding by identity tuple

jhrozek:
## Summary

- Replaces the vMCP HMAC-of-bearer-token session binding with a stable `(iss, sub)` identity tuple so legitimate OAuth refreshes no longer terminate sessions (users currently get logged out once per access-token TTL).
- Drops the per-deployment HMAC secret and the operator-side plumbing that distributed it.
- Closes a prior `LocalUserMiddleware` gap where every local user fell into the anonymous equivalence class.
- Supersedes the session-binding portion of THV-0038; preserves the cross-pod persistence design of THV-0047 with the new key.

Tracks [toolhive#5306](https://github.com/stacklok/toolhive/issues/5306).

## Review focus

- Threat model in §Security Considerations — particularly the HMAC → plaintext-PII-at-rest trade-off and the operator mitigations.
- OAuth/OIDC correctness (refresh invariance, `(iss, sub)` global uniqueness, RFC 8693 interactions).
- Migration story (invalidate-on-read) and the rolling-deploy race note.

### DIFF
--- a/rfcs/THV-0073-vmcp-session-identity-binding.md
+++ b/rfcs/THV-0073-vmcp-session-identity-binding.md
@@ -206,6 +206,16 @@ If a future incoming-auth type uses RFC 7662 token introspection (it currently d
 - No public API change: the affected packages are all internal.
 - Legacy metadata keys remain defined for one release cycle so the migration's legacy-detection branch can read them; they are removed in the cleanup phase.
 
+### Rollback
+
+Sessions written by the new code contain only `MetadataKeyIdentityBinding` and not the legacy `MetadataKeyTokenHash` / `MetadataKeyTokenSalt`. When the prior image's session manager loads such a session, `loadSession` checks for `MetadataKeyTokenHash` before delegating to `RestoreSession`. Because the key is absent, `loadSession` returns `transportsession.ErrSessionNotFound` immediately — the session is never passed to `RestoreSession`, and the hijack-prevention decorator is never constructed without a binding. A second independent fail-closed gate inside `RestoreSession` catches any path that might somehow reach it. There is no code path that restores a session without the decorator. The client receives the standard MCP re-init signal and starts a fresh session — the same operator-visible cost as the forward-migration path.
+
+Operator consequence: rolling back to the prior image invalidates new-format sessions (one re-auth per active session) but the system is safe. There is no need to wait out the session TTL before rolling back.
+
+A defensive write of a sentinel `MetadataKeyTokenHash` from the new code is explicitly rejected. An empty sentinel would cause the prior image's `RestoreHijackPrevention` to treat the session as anonymous (`allowAnonymous=true`) — a security downgrade worse than a clean re-init.
+
+The prior image's `Terminate` uses `MetadataKeyTokenHash` presence as the Phase-1/Phase-2 discriminator. On rollback, terminating a new-format session takes the Phase-1 placeholder path (marks `terminated=true` rather than immediately deleting it), so the session lingers in Redis until its TTL expires. This is a behavioural oddity, not a security regression — both `Validate` and `loadSession` reject terminated sessions — but operators should expect rolled-back-and-terminated sessions to remain visible in Redis until TTL.
+
 ### Forward compatibility
 
 - RFC 7662 introspection-based incoming auth would require the IdP to emit `iss` and `sub` in introspection responses. A startup probe that fails fast on incompatible IdPs is recommended when that mode is added — it's not a hard precondition because the failure mode is loud (extraction errors on every request) and surfaces immediately in operator logs.
--- a/rfcs/THV-0073-vmcp-session-identity-binding.md
+++ b/rfcs/THV-0073-vmcp-session-identity-binding.md
@@ -1,0 +1,292 @@
+# RFC-0073: vMCP Session Binding by Identity Tuple
+
+- **Status**: Draft
+- **Author(s)**: Jakub Hrozek (@jhrozek)
+- **Created**: 2026-05-20
+- **Last Updated**: 2026-05-21
+- **Target Repository**: toolhive
+- **Related Issues**: [toolhive#5306](https://github.com/stacklok/toolhive/issues/5306)
+- **Supersedes**: the session-binding portion of [THV-0038](./THV-0038-session-scoped-client-lifecycle.md); updates [THV-0047](./THV-0047-vmcp-proxyrunner-horizontal-scaling.md) to use the new binding format
+
+## Summary
+
+The Virtual MCP server (vMCP) currently binds each MCP transport session to an HMAC-SHA256 hash of the raw bearer-token bytes presented at session creation, then rejects any subsequent request whose token bytes hash differently. This terminates the session on every legitimate OAuth refresh because the new access token has different bytes for the same identity. This RFC replaces the binding with a stable `(iss, sub)` identity tuple extracted from the OIDC identity's `Claims`, eliminates the HMAC plumbing entirely, and migrates legacy sessions via an invalidate-on-read scheme.
+
+## Problem Statement
+
+[THV-0038](./THV-0038-session-scoped-client-lifecycle.md) (Session-Scoped Architecture for vMCP) introduced session-hijack prevention with the requirement:
+
+> "Store a secure cryptographic hash of the original authentication token in the session during creation. To prevent offline attacks if session state is leaked (e.g., from Redis/Valkey), prefer a keyed hash (e.g., `HMAC-SHA256` with a server-managed secret and a per-session salt). [...] On each request, validate the current auth token against the session's bound hash using a constant-time comparison to prevent timing attacks. If mismatch, reject with 'session authentication mismatch' error and terminate session."
+
+[THV-0047](./THV-0047-vmcp-proxyrunner-horizontal-scaling.md) (Horizontal Scaling) operationalised this: the `HijackPreventionDecorator` persists `boundTokenHash` and `tokenSalt` in `MetadataKeyTokenHash` / `MetadataKeyTokenSalt` so the binding survives cross-pod restore.
+
+Both RFCs are correct about the threat they defend — a stolen `Mcp-Session-Id` paired with the attacker's own token must not be accepted. Both are silent on one specific operation: **OAuth `refresh_token` exchange**. RFC 6749 §6 produces a new access token with different bytes for the same End-User. Under the HMAC-byte binding, every refresh fails the hijack check and the user is logged out — typically once per access-token TTL window (e.g. every 5 minutes for short-lived ATs).
+
+The bug surfaces as:
+
+```
+HTTP 401 / JSON-RPC isError=true
+  body:  "Unauthorized: caller identity does not match session owner"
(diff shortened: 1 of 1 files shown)


rfcs/THV-0073-vmcp-session-identity-binding.md
```diff
@@ -0,0 +1,292 @@
...
+- TLS for Redis transport when not on a confined network.
+- Treat session-store dumps as an identity-data incident, not opaque-token loss.
+
+### OAuth and OIDC correctness
+
+The binding deliberately uses `(iss, sub)` and *only* `(iss, sub)`:
+
+- **Refresh invariance**: OIDC Core §2 requires `sub` to be stable per End-User within an issuer; RFC 6749 refresh does not change the principal. Bindings survive refresh.
+- **Global uniqueness**: `(iss, sub)` is globally unique under OIDC Core §2 — `sub` is only locally unique within `iss`, but the `iss\x00sub` concatenation lifts that to global.
+- **Token exchange (RFC 8693)**: The binding is at the front-door token, not at any upstream-exchanged token. RFC 8693 §4.1 impersonation preserves `sub`; delegation adds `act` but `sub` still describes the principal the token is *about*. No interaction with the binding.
```
tgrunnagle, line 150: Question: for future proofing, should we also bind the session to an optional `act` claim? If different agents are accessing the vMCP with a delegated token for the same user we should differentiate them.

If so, do we need to consider nested delegation?

tgrunnagle (dismissed):
LGTM, one non-blocking question

rfcs/THV-0073-vmcp-session-identity-binding.md
```diff
@@ -0,0 +1,292 @@
...
+- **Why not `aud`**: step-up auth and audience-scoped tokens for different upstreams must not force a new vMCP session. `aud` validation belongs in the token-validation layer, one above the session boundary.
+- **Why not `acr`/`amr`**: step-up enforcement is Cedar's job (authz layer), not the session layer.
+- **Why not `act`**: describes who is using the token, not who it is about — wrong axis for session identity.
+
+### AnonymousMiddleware scoping
+
+`AnonymousMiddleware` emits identical `(iss="toolhive-local", sub="anonymous")` for every request. Consequently every anonymous user collapses into one binding equivalence class — per-identity hijack prevention is impossible *by construction* in anonymous deployments. This is not a regression: the old token-hash scheme produced the empty-string sentinel for the same case and exhibited identical behaviour. We surface the limitation at startup with a `WARN` and document the mode as dev-only. The session decorator still defends against the *upgrade attack* — a token presented to an anonymous session is rejected.
```
tgrunnagle: **HIGH — Local user named "anonymous" collides exactly with the anonymous-mode sentinel binding** (Consensus 9/10) — *raised by: oauth-oidc, security, mcp-protocol, migration-ops*

`AnonymousMiddleware` emits `(iss="toolhive-local", sub="anonymous")` (this line). The `LocalUserMiddleware` fix at line 161 says the middleware "populates with the username as `sub`" — so a local user literally named `anonymous` produces an identical `(iss="toolhive-local", sub="anonymous")` binding. A Redis dump containing `toolhive-local\x00anonymous` becomes ambiguous (anonymous-mode session vs local user "anonymous"), and the LocalUserMiddleware fix the RFC claims as a "strict improvement" silently regresses for that one username.

**Recommendation**: Either reserve `anonymous` as a forbidden local username in `LocalUserMiddleware`'s validation (with a clear startup error), or use distinct `iss` values for the two modes — e.g. `iss="toolhive-anonymous"` for anonymous mode vs `iss="toolhive-local"` for LocalUserMiddleware. Option 2 removes the collision by construction. Document the chosen approach in the Fail-closed defaults section.

tgrunnagle (commented):
## Multi-Agent Consensus Review — HIGH-severity findings (non-blocking)

Agents consulted: security & threat model, OAuth/OIDC, MCP protocol, migration & ops, API design, RFC structure, general code quality.

The RFC's approach is sound: it correctly identifies the bug in toolhive\#5306 as binding the wrong invariant, correctly chooses `(iss, sub)` over the rejected two-tier `tsid + iss/sub` design, correctly rejects the hybrid migration that would re-enable the hijack attack, and correctly preserves THV-0047's cross-pod persistence model with only the persisted value changed. The threat-model framing ("Both defend / New defends, old did not / Old defended, new does not") is unusually clear, and the RFC is faithful to the panel decisions documented in the linked issue.

A multi-agent review surfaced two HIGH-severity items worth resolving before implementation begins. **These are not merge blockers** — both are design questions that can be answered in this RFC iteration, and the rest of the design stands. The detailed inline comments below describe each issue and a concrete recommendation. A larger set of MEDIUM-severity items (global-uniqueness assumption, restore-branch-3 error class, `UnauthenticatedSentinel` rationale, refresh-invariance citation cleanup, session-affinity portability, anonymous-mode production guardrail, missing template sections, etc.) was identified locally but is not posted here — happy to share that fuller set if useful.

---
*Generated with [Claude Code](https://claude.com/claude-code)*

rfcs/THV-0073-vmcp-session-identity-binding.md
```diff
@@ -0,0 +1,292 @@
...
+
+**Encrypted-at-rest binding.** An application-layer encryption secret colocated with vMCP pods has the same leak surface as the binding it protects. Layered defence belongs at the store (ACLs, NetworkPolicy), not at the application — this is the same class of misplaced complexity as the original HMAC-secret approach.
+
+**Hybrid migration ("read both, write new").** Re-creates the exact attack the binding defends against: an attacker presenting a stolen legacy session-id plus their own bearer token would cause the session to adopt the *attacker's* identity on the rebind. Invalidate-on-read fails closed at the cost of one re-auth at deploy.
+
+**Keep the HMAC scheme, hash `(iss, sub)` instead of token bytes.** Survives refresh and preserves at-rest opacity, but the HMAC adds operational complexity (per-deploy secret, cross-replica coordination) for a property that is *not in the threat model* — the binding is not a credential, so identifying the user from a Redis dump does not enable impersonation. Removing the HMAC produces simpler, more honest code and reduces operator-side complexity (Goal 4).
+
+## Compatibility
+
+### Backward compatibility
```
tgrunnagle, line 202: **HIGH — Rollback behaviour for new-format sessions hitting the old image is undefined** (Consensus 8/10) — *raised by: migration-ops*

The Backward Compatibility section documents forward migration (legacy sessions hitting new code → invalidate-on-read), but is silent on the reverse: sessions written with only `MetadataKeyIdentityBinding` and no legacy keys are read by the old image, which looks for `MetadataKeyTokenHash` / `MetadataKeyTokenSalt`. Depending on how legacy `RestoreHijackPrevention` handles "metadata present but expected keys absent", this may surface as HTTP 500, or — worse — as a session restored *without* any hijack-prevention decorator applied (silent security regression). Operators considering Phase-1 rollback need an explicit behavioural guarantee.

**Recommendation**: Add a Rollback subsection that (a) states exactly how the prior image behaves when it encounters a session with only the new key, (b) either commits to a defensive write of a sentinel legacy key during Phase 1 or explicitly documents that rollback requires waiting out the session TTL, and (c) adds a rollback test mirroring the rolling-deploy race test (new-image-written session read by old factory).

rfcs/THV-0073-vmcp-session-identity-binding.md
jhrozek: Looked at this more carefully — sticking with `(iss, sub)` only, but rewriting the rejection in the RFC since the one-liner was too thin.

The "wrong axis" framing is actually wrong — session-hijack defence IS an actor-axis concern. The real reason `act` doesn't help is that the captured `Mcp-Session-Id` isn't a credential. Anyone who can present a same-`sub` token can open their own session, so the captured id buys no privilege escalation under the current feature set. And binding to `act` would break legitimate agent-credential rotation — same class of bug we just fixed.

Cedar deferral also gets tightened: the nested-claim conversion isn't in the codebase yet, it's planned in `THV-XXXX-user-to-agent-delegation`. Reworded so we don't pretend it works today.

Adding a forward-compat clause: any future RFC introducing session-pinned state (elicitation consents, long-lived tool grants) has to explicitly revisit this binding rather than inherit `(iss, sub)` by default. That's what keeps the deferral from quietly becoming a permanent gap.

Nested delegation per RFC 8693 §4.1: chain depth is informational only for access control, so the binding stays single-level.

Pushed in 9c70578.

rfcs/THV-0073-vmcp-session-identity-binding.md
jhrozek: Confirmed — going with your option 2 (distinct `iss` per mode), eliminates the collision by construction.

`AnonymousMiddleware` switches to `iss="toolhive-anonymous"`; `LocalUserMiddleware` keeps `iss="toolhive-local"`. A local user named `anonymous` now produces `toolhive-local\x00anonymous`, structurally distinct from anonymous-mode's `toolhive-anonymous\x00anonymous`.

One literal change in `pkg/auth/anonymous.go` plus the test, no other call sites depend on the string. Adding a `toolhive-` prefix convention to forward-compat so future internal middleware can't reintroduce the class.

Adjacent note: `UnauthenticatedSentinel = "unauthenticated"` would theoretically collide if anyone names a local user `unauthenticated`, but it's only used as a session-store value, not a `sub` claim, so not exploitable today.

Pushed in f85f502.

rfcs/THV-0073-vmcp-session-identity-binding.md
jhrozek: Checked the code on `main` — gap is doc-only, not a code deficiency.

`loadSession` in `pkg/vmcp/server/sessionmanager/session_manager.go` checks `MetadataKeyTokenHash` before calling `factory.RestoreSession`. New-format sessions (only `MetadataKeyIdentityBinding`) trip this and return `ErrSessionNotFound` — the session never reaches `RestoreSession`, and the decorator is never constructed without a binding. A second independent gate inside `RestoreSession` catches anything that might. No HTTP 500, no silent decorator-less restore. Operator-visible result on rollback: one re-auth per active session, same cost as the forward migration.

Going with (a) [doc-only], adding (c) [rollback regression test] for a follow-up. Skipping (b) — a defensive sentinel write of an empty token-hash would actually be harmful: the prior image's `RestoreHijackPrevention` treats empty as `allowAnonymous=true`, which is worse than a clean re-init. Adding explicit rejection of that approach.

One extra oddity worth flagging: `Terminate` on the prior image uses `MetadataKeyTokenHash` presence as the Phase-1/Phase-2 discriminator, so terminating a new-format session on the rolled-back image marks it `terminated=true` instead of deleting it. Lingers in Redis until TTL. Not a regression (both `Validate` and `loadSession` reject terminated sessions), but operator-visible.

Pushed in 97613f4.

jerm-dro (approved):
Thanks for writing this up. I appreciate the clarity on the threat model impact. This is a reasonable simplification.


FYI: The next iteration of MCP gets away with Sessions entirely: https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

so we can eventually delete this session storage complexity. In the meantime, making things functional SGTM.

PR #76, opened 2026-06-02
RFC: vMCP Core Interface — Splitting Domain Logic from Transport

tgrunnagle:
## Summary

Proposes extracting a small, identity-parameterized `VMCP` domain interface from the Virtual MCP Server, paired with `New(cfg) -> VMCP` to construct the core and `Serve(ctx, VMCP, serverCfg) -> *Server` to stand up the transport. Embedders can wrap the core with their own decorators to augment behavior — without reimplementing aggregation/routing/composition, and without coupling to the mcp-go SDK lifecycle, the HTTP middleware chain, or request-context contracts. The existing `server.New(...)` entry point is preserved as a thin wrapper over `Serve(ctx, New(cfg), serverCfg)`, so current embedders are unaffected.

## Key points

- Minimal `VMCP` interface exposing the MCP capability surface (list/call tools, list/read resources, list/get prompts, lookup) with identity as an **explicit parameter** — removing the `auth.IdentityFromContext` coupling on the domain path.
- `New` relocates (not rewrites) the collaborator wiring `server.New` does today; `Serve` owns all transport concerns (mcp-go, middleware chain, embedded AS, status, telemetry, health).
- Decoration is the supported domain-extension mechanism; list and call share one source of truth so a decorator cannot drift or widen access.
- No change to serialized config, CRD/YAML, data model, or incoming/outgoing auth strategies. `server.New` signature stays stable throughout; the split lands behind it across small, sequenced PRs.

### DIFF
--- a/rfcs/THV-0076-vmcp-core-interface.md
+++ b/rfcs/THV-0076-vmcp-core-interface.md
@@ -1,0 +1,694 @@
+# RFC-0076: vMCP Core Interface — Splitting Domain Logic from Transport
+
+- **Status**: Draft
+- **Author(s)**: Trey Grunnagle (@tgrunnagle)
+- **Created**: 2026-06-01
+- **Last Updated**: 2026-06-01
+- **Target Repository**: toolhive
+- **Related Issues**: _(none yet)_
+
+## Summary
+
+This RFC proposes extracting a small, identity-parameterized `VMCP` domain
+interface from the Virtual MCP Server, paired with a `New(cfg) -> VMCP`
+constructor for the core and a `Serve(ctx, VMCP, serverCfg) -> *Server` helper
+for the transport layer. Embedders can then wrap the core with their own
+decorators to augment behavior — without reimplementing aggregation, routing,
+or composition, and without coupling to the mcp-go SDK lifecycle, the HTTP
+middleware chain, or request-context contracts. The existing
+`server.New(...)` entry point is preserved as a thin wrapper over
+`Serve(ctx, New(cfg), serverCfg)` so current embedders are unaffected.
+
+## Problem Statement
+
+The `pkg/vmcp/` packages are designed as an embeddable Go library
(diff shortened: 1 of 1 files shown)


rfcs/THV-0076-vmcp-core-interface.md
```diff
@@ -0,0 +1,694 @@
...
+`v.ListTools(ctx, identity)` (and friends) directly, removing the
+context-coupling described in [anti-pattern #1](https://github.com/stacklok/toolhive/blob/main/.claude/rules/vmcp-anti-patterns.md#1-context-variable-coupling).
+
+The core `VMCP` is **stateless with respect to sessions**: it aggregates on
+demand and holds no per-session state. `Serve` is therefore responsible for
+everything session-scoped. It captures the aggregated capability set once per
+MCP session (it already owns the session lifecycle) and reuses it for the life
+of that session, preserving today's "capabilities fixed at `initialize`"
+consistency guarantee without putting a cache in the core (consistent with
+anti-patterns
```
jerm-dro, line 219: I'm curious how this shakes out. The `MultiSession` is session-aware, but maybe that remains an implementation detail of the core `VMCP`. Or does the session concept get hoisted into the server? Your proposal might make more sense. Otherwise, you'd need interface methods on the `VMCP` object to return the session ID, which we don't want. The session concept is going away in subsequent versions of MCP, so this should remain an implementation detail either way.

jerm-dro (approved):
Looks good! Let 'er rip!

JAORMX (approved):

rfcs/THV-0076-vmcp-core-interface.md
tgrunnagle: I'd like to extract all session-related components into the Server layer, but need to dig into what refactoring `MultiSession` would look like. I'll be sure to flesh that out during implementation planning.

PR #77, opened 2026-06-15
RFC: Plugin lifecycle management in ToolHive

JAORMX:
## Summary

Proposes adding **plugin** lifecycle management to ToolHive, mirroring the existing skills system ([RFC-0030](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0030-skills-lifecycle-management.md)). A *plugin* is the cross-vendor "bundle of primitives" unit (slash commands, subagents, Agent Skills, hooks, MCP server configs, LSP servers) declared by `.claude-plugin/plugin.json`.

ToolHive will let users **build** a plugin directory into a reproducible OCI artifact, **push** it to any OCI registry, **install** it (from registry name / OCI ref / `git://`), and **list/info/uninstall** it — reusing the registry, OCI, groups, and SQLite `entries` infrastructure that already serves skills. As a bridge to native client tooling, it can also generate a `marketplace.json`.

## Why

- Plugins are the industry-convergent bundle unit (Claude Code, Cursor, Codex, Copilot, Gemini, Kiro), but **no one packages a multi-primitive plugin bundle as an immutable, content-addressable, signable OCI artifact** — an open niche ToolHive is positioned to fill.
- The skills system was deliberately built on a generic foundation (`entry_type` discriminator, `x/dev.toolhive/<type>` registry namespace, shared OCI primitives) that anticipates exactly this extension.

## What's genuinely new vs. skills

1. **Install target** — plugins normally need marketplace-cache + `settings.json` mutation. The RFC recommends the in-place *skills-directory-plugin* mechanism for v1 (pure filesystem, no settings mutation), with marketplace-cache deferred as a per-client `PathResolver` strategy. (Open Question #1)
2. **Executable surface** — unlike inert skills, plugins bundle hooks and MCP servers (code). The security section centers on this: a pre-install executable-surface inventory in `thv plugin info`, `--require-signature` via cosign + the OCI Referrers API, and install-time audit events.

## Notes for reviewers

- File is named `THV-XXXX-...` per the RFC convention; will rename to match this PR number.
- Touches `toolhive` (primary), `toolhive-core` (new `oci/plugins` + shared-primitive refactor), and `toolhive-registry-server` (plugin catalog).
- Key product decisions to weigh: Open Question #1 (install target) and #4 (running bundled MCP servers through `thv` with isolation as the v2 headline).


aponcedeleonch (approved):
Agreed on the call. Native-bundle materialization for v1, fan-out deferred to a future adapter. And your two structural points land: fan-out can't add a capability a client lacks, and the `${CLAUDE_PLUGIN_ROOT}` constraint argues for keeping the tree intact, not against it. I'll concede both.

I did go and check what each client's bundle format actually is, since "shipped a native bundle format" is doing a lot of work in the ~8-of-10 number, and the formats turn out not to be the same kind of thing. Three tiers:

Near drop-in, and the reason matters: Cursor, Copilot, and Codex are cheap to target specifically because they converged on the `.claude-plugin` format. Cursor mirrors the component dirs and `SKILL.md` layout almost exactly. Copilot literally accepts `.claude-plugin/plugin.json` and ships a `${PLUGIN_ROOT}` variable. Codex accepts both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` and even aliases `CLAUDE_PLUGIN_ROOT`. That convergence is the actual strength of the native-bundle thesis, and I'd lead the doc with it rather than with a client count.

Real bundle, but a genuine per-client adapter: Gemini CLI has full primitive parity but its own `gemini-extension.json` manifest and TOML command files, so commands need transcoding. Kiro Powers are a real multi-primitive bundle but the manifest is a Markdown `POWER.md` with no commands/agents/skills split. Neither is "place the tree"; both are mapping work we'd need to scope.

Not an installable plugin bundle at all: Goose "recipes" are run-on-invocation workflow files (a parameterized prompt plus inline MCP), not a set of primitives you install. Continue "bundles" are explicitly a hub-side UI grouping that, per their own docs, isn't represented in config files; "add bundle" just expands into individual block references.

That last tier is where I'd push back on the "~1.5 fan-out targets" number. Goose and Continue are in the bundle column in your count, but you can't materialize a bundle into either. They're fan-out-or-skip clients, same as OpenCode and Zed. So the real fan-out-relevant set is closer to four than to one and a half.

None of this changes the v1 decision. Defer fan-out, ship native bundle, and the `.claude-plugin` convergence makes that genuinely cheap for the three clients that matter most. I'd just adjust the doc to (1) rest the thesis on that convergence rather than an 8-of-10 reach, (2) name the Gemini and Kiro adapter cost explicitly instead of folding them into "native bundle just works," and (3) be upfront that Goose and Continue are fan-out-or-skip, not bundle-place.

Approving on that basis.

jhrozek (commented):
adding a few comments as I read, none blocking. the two I'd actually want to settle before merge are the materialization target (in-place skills-dir vs the native staging your top comment proposes) and the matching mcp question at the bottom, since the doc and your own comment point opposite ways there. the rest are nits / a couple factual fixes on the codex section.

rfcs/THV-0077-plugins-lifecycle-management.md
```diff
@@ -0,0 +1,730 @@
...
+
+- **v1 (recommended): in-place skills-directory plugin.** ToolHive extracts the plugin tree to `<client-skills-dir>/<name>/` (e.g. `~/.claude/skills/<name>/` for user scope). Claude Code auto-loads **all components** next session. No `settings.json` mutation, the install/uninstall lifecycle is pure filesystem (mirroring skills exactly), and `${CLAUDE_PLUGIN_ROOT}` resolves correctly because the client computes it from the load location.
```
jhrozek, line 176: honest question, mostly to make sure we've actually landed this — your comment up top argues for dropping the in-place skills-dir approach and staging into a thv-managed dir registered as a local-path marketplace, but the doc still recommends in-place skills-dir for v1. which one are we going with?

the skills-dir path also shares `~/.claude/skills/` with `thv skill` — both extract into the same tree and the only thing telling a plugin from a skill is whether there's a `.claude-plugin/plugin.json`, so `thv skill list` and `thv plugin list` end up scanning the same dir. I'd rather pin this down before phase 3 than carry it as open question #1a.

rfcs/THV-0077-plugins-lifecycle-management.md
```diff
@@ -0,0 +1,730 @@
...
+
+For completeness, Claude Code also supports two no-settings, **session-only** mechanisms that load all components — `claude --plugin-dir ./path` and `claude --plugin-url <zip>` — and a persistent local-marketplace path (`/plugin marketplace add ./path`) that loads everything but *does* mutate settings (`extraKnownMarketplaces`). These are useful for `thv plugin` dev/test workflows but are not the managed-install target.
+
+#### Codex: a second v1 adapter, with a heavier mechanism and reduced fidelity
+
+The same packaged `.claude-plugin/` tree is *readable* by Codex — its loader supports `.claude-plugin/plugin.json` as a fallback manifest (`ALTERNATE_PLUGIN_MANIFEST_RELATIVE_PATH`), so no separate manifest is required. But — **verified against `openai/codex` source** — Codex's materialization is fundamentally different from Claude Code's, in two ways that the adapter must handle and that `thv plugin install` must surface to the user:
```
jhrozek, line 185: two small things here, both so nobody bounces off it checking the source:

- `ALTERNATE_PLUGIN_MANIFEST_RELATIVE_PATH` only exists in the `#[cfg(test)]` blocks. the shipping constant is `DISCOVERABLE_PLUGIN_MANIFEST_PATHS` in `codex-rs/utils/plugins/src/plugin_namespace.rs`. the fallback's real, just cite the one that's actually compiled in.
- a few lines down (L190), "not feature-flag-gated" isn't right — it's gated by `Feature::Plugins`, it's just `Stage::Stable, default_enabled: true`. the off-by-default one is `Feature::RemotePlugin` (the remote catalog). conclusion holds, but the wording reads as false to anyone grepping `features/src/lib.rs`.

rfcs/THV-0077-plugins-lifecycle-management.md
```diff
@@ -0,0 +1,730 @@
...
+```
+thv plugin install <source> [flags]
+  --clients strings      Target client(s) with a materialization adapter
+                         (v1: claude-code, codex; default: adapter-supported clients on host)
+  --scope string         user | project (default "user")
+  --project-root string  Project root for project scope
+  --group string         Add plugin to a group
+  --version string       Version/tag to resolve for a plain name
+  --force                Overwrite an existing unmanaged plugin
+  --require-signature    Refuse to install unless a valid signature is present
```
jhrozek, line 319: same thing I asked on CIMD — do we want this opt-in? the security section says installing a plugin is granting code execution, and then leaving unsigned-by-default as the easy path feels backwards. not asking for signature-required in v1 necessarily, but could the default be a loud confirm when there's hooks/mcp and no signature, and keep `--require-signature` as the hard gate for CI? opt-out for the scary case reads safer than opt-in.

rfcs/THV-0077-plugins-lifecycle-management.md
```diff
@@ -0,0 +1,730 @@
...
+
+- v1 logs structured events from the CLI/API service layer: `plugin_installed`, `plugin_uninstalled`, `plugin_upgraded`, `plugin_built`, `plugin_pushed`, each with name, digest, source, signature status, and the component inventory (notably hook count). This gives a fleet a queryable inventory of *what executable plugin surface is installed where*.
+- A full audit-architecture for client-side artifacts remains the same separate-RFC concern flagged in RFC-0030, but the install-time event log is in scope here because the executable surface makes "who installed what hook" a real security need.
+
+### Mitigations
+
+| Threat | Mitigation (v1) | Future |
+|--------|-----------------|--------|
+| Malicious hook / MCP server | `thv plugin info` surfaces the **executable-surface inventory** (hooks + MCP commands) before install; install-time event logging | policy: refuse hooks not on an allowlist; org policy profiles |
+| Trojaned update | digest recorded; upgrades show digest change; component-inventory diff on upgrade | require re-approval when hook set changes |
```
jhrozek, line 562: I'd pull this into v1 rather than leave it in the Future column. we're already recording the component inventory and the digest, so an upgrade quietly adding a hook is exactly what that inventory is there to catch. deny-by-default when the hook/mcp set changes, with a `--accept-new-executables` to proceed, seems cheap given the data's already on hand.

rfcs/THV-0077-plugins-lifecycle-management.md
```diff
@@ -0,0 +1,730 @@
...
+2. **Component-inventory diff on upgrade.** Should an upgrade that *adds* a hook or MCP server require explicit re-approval (`--accept-new-executables`) by default?
+3. **`userConfig` / sensitive options.** Plugins can declare config (some `sensitive`). Does ToolHive stay entirely hands-off (client keychain owns it), or offer to wire values from ToolHive's secrets manager at install time?
+4. **Managed MCP via dependency references (deferred — follow-up RFC).** The intended model is for a plugin to *reference* first-class ToolHive-managed MCP servers via `requires` rather than bundle executable `command`/`args`. Open: does `requires` carry OCI image references, registry names, or both? How are referenced servers resolved to workloads and wired into the client config at the proxy URL? What default permission profile applies to a server pulled in by a plugin? (The alternative — repackaging a bundle's local `command`/`args` into a container, or adding a host-process runtime — was considered and set aside in favor of references; see Alternative 7.)
```
jhrozek, line 693: flip side of my skills-dir question above: the doc defers bundled mcp to the `requires`-reference model, but your top comment frames running the bundled servers through the runtime as the actual v1 thesis (the one thing a `git clone` doesn't give you). I'm fine with either honestly, the reference model is cleaner long term — but the comment and the doc point opposite ways and the thread never came back to it. can we mark which one won? would rather not merge an approved doc that argues with its own author.

jhrozek (approved):
solid RFC, really thorough research behind it. left a few comments inline but nothing blocking — the only two I'd genuinely like settled are the materialization target and the matching mcp question, since the doc and your top comment point opposite ways there. happy for those to be a follow-up. approving.

rfcs/THV-0077-plugins-lifecycle-management.md
JAORMX: Good catch — the RFC was inconsistent with the direction we had converged on. I updated it to make the v1 Claude Code adapter use a ToolHive-managed local marketplace + native plugin cache/settings path, not the skills-dir path. The skills-dir mechanism is now only documented as an alternative not chosen, because it overlaps with `thv skill` state, gives installs a special `@skills-dir` identity, and has confusing project-scope behavior.

rfcs/THV-0077-plugins-lifecycle-management.md
JAORMX: Fixed both. The fallback now cites `DISCOVERABLE_PLUGIN_MANIFEST_PATHS` instead of the test-only `ALTERNATE_PLUGIN_MANIFEST_RELATIVE_PATH`. I also changed the feature-gate wording: plugin loading is behind `Feature::Plugins`, but that feature is `Stage::Stable` and `default_enabled: true`; the off-by-default pre-GA bit is `Feature::RemotePlugin`.

rfcs/THV-0077-plugins-lifecycle-management.md
JAORMX: Agreed. I added this as v1 behavior: if a plugin is unsigned and declares hooks or MCP servers, interactive installs show the executable-surface inventory and require explicit confirmation. Non-interactive installs fail unless `--accept-unsigned-executables` is set. `--require-signature` remains the hard CI/policy gate.

rfcs/THV-0077-plugins-lifecycle-management.md
JAORMX: Agreed and moved into v1. The RFC now says upgrades diff the recorded component inventory, and adding hooks or MCP servers is denied by default unless the user passes `--accept-new-executables`.

rfcs/THV-0077-plugins-lifecycle-management.md
JAORMX: Yes — the dependency-reference model won. I tightened the RFC so bundled MCP remains explicitly unmanaged in v1, and the planned managed path is `requires` references to first-class ToolHive-managed MCP servers. Running bundled local `command`/`args` through ToolHive stays in Alternative 7 / follow-up RFC territory, not v1.

PR #79, opened 2026-06-26
RFC: Identity-Assertion JWT Authorization Grant (ID-JAG / cross-app access)

jhrozek:
## Summary

Adds an RFC proposing first-class support for the IETF OAuth WG draft *Identity Assertion JWT Authorization Grant* (`draft-ietf-oauth-identity-assertion-authz-grant`, ID-JAG) — also known as **cross-application access (XAA)** — in ToolHive's Virtual MCP Server.

ID-JAG composes RFC 8693 token exchange (Step A, at the front-door IdP) and RFC 7523 JWT bearer (Step B, at the backend's resource AS) to materialise a per-backend access token from the user's upstream ID token, with no extra browser interaction. It closes the cross-domain trust gap that the existing `token_exchange` and `upstream_inject` strategies cannot cover (the "eager-chaining" problem).

## What the RFC proposes

Builds on foundation already merged to `toolhive/main` (the `pkg/oauthproto/jwtbearer` RFC 7523 client and the `tokenexchange` `RequestedTokenType`/`Resource` fields) and adds:

- `Identity.UpstreamIDTokens` plumbing so the auth middleware surfaces the upstream ID token to outgoing strategies;
- a new `xaa` outgoing-auth strategy performing the two-step ID-JAG flow end to end;
- operator/CRD support via an `xaa` `MCPExternalAuthConfig` type.

PoC-validated against the xaa.dev sandbox. Strictly additive — existing strategies are untouched.

## Notes for reviewers

- File uses the `THV-XXXX` placeholder; it will be renamed to this PR's number per the [naming convention](https://github.com/stacklok/toolhive-rfcs/blob/main/CONTRIBUTING.md#rfc-file-naming-format).
- Security section documents two known limitations shipping as-is: audit logging on the outgoing-auth path is net-new wiring (§4.7), and client-secret rotation requires a proxy-runner restart (§4.8).

Related: stacklok/toolhive#5218


JAORMX (dismissed):
Really appreciate the background info at the beginning of the RFC, helped me get up to speed on this.

rfcs/THV-0079-cross-app-access-grant.md
```diff
@@ -0,0 +1,682 @@
...
+    TokenTypeAccessToken = "urn:ietf:params:oauth:token-type:access_token"
+    TokenTypeIDToken     = "urn:ietf:params:oauth:token-type:id_token"
+    TokenTypeJWT         = "urn:ietf:params:oauth:token-type:jwt"
+    TokenTypeIDJAG       = "urn:ietf:params:oauth:token-type:id-jag"
+)
+
+// Grant type URNs.
+const (
+    GrantTypeTokenExchange = "urn:ietf:params:oauth:grant-type:token-exchange"
+    GrantTypeJWTBearer     = "urn:ietf:params:oauth:grant-type:jwt-bearer"
```
tgrunnagle, line 227: **[F1, 9/10] 🔴 HIGH: `TokenTypeIDJAG` URN may be missing `+jwt` suffix**

The constant is defined as `"urn:ietf:params:oauth:token-type:id-jag"`, but the JWT header `typ` for the returned assertion is `"oauth-id-jag+jwt"`. By the IETF typed-JWT convention (as in `at+JWT` for JWT access tokens), the token type URN should carry the full `typ` value as suffix: `urn:ietf:params:oauth:token-type:id-jag+jwt`. If the draft registers this as the canonical URN, Step A will send an incorrect `requested_token_type` to any conformant IdP and the `issued_token_type` check in §3.6 will never match — the two-step flow fails against any conformant IdP.

Please verify against draft-ietf-oauth-identity-assertion-authz-grant-04 §3 / IANA registration. If `+jwt` is required, also update the Step A example request at line 70.

rfcs/THV-0079-cross-app-access-grant.md
```diff
@@ -0,0 +1,682 @@
...
+
+`SubjectProviderName` is the only field in `XAAConfig` without an equivalent in any other open-source ID-JAG implementation. All other implementations pass the subject token directly; ToolHive's multi-upstream architecture requires a *selector* that identifies which of the configured upstream providers' ID tokens to use as the Step A subject. With a single upstream — the common case — the operator can omit the field and rely on the default.
+
+### 3.8 Operator and CRD Integration
+
+The CRD type lives in `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go`. It adds the type constant `ExternalAuthTypeXAA = "xaa"` and a spec struct `XAASpec` carrying the same Step A / Step B fields as `XAAConfig`, with secret references (`IDPClientSecretRef`, `TargetClientSecretRef`) instead of inline secrets. The value `"xaa"` is intentionally identical on both the snake_case vMCP strategy surface (`StrategyTypeXAA`) and the camelCase CRD surface (`ExternalAuthTypeXAA`); as an acronym it carries no compound to case-fold, matching the existing `ExternalAuthTypeOBO = "obo"` precedent.
```
tgrunnagle, line 358: **[F9, 7/10] `SubjectProviderName` defaulting undefined for multi-upstream deployments**

With a single upstream provider this is unambiguous. With two or more providers configured in `authServerConfig.upstreamProviders` and `SubjectProviderName` omitted, there is no documented selection rule — the RFC says "the common case" is one provider but gives no deterministic choice or error path for the multi-upstream case. A silently-wrong provider selection produces a confusing Step A failure at the IdP.

Suggest specifying: when `SubjectProviderName` is omitted and multiple upstream providers are configured, startup validation must return an explicit error requiring the field. This is safer than any implicit selection rule for a security-relevant config.

rfcs/THV-0079-cross-app-access-grant.md
```diff
@@ -0,0 +1,682 @@
...
+
+## 4. Security Considerations
+
+ID-JAG is a privilege-relevant operation: vMCP asks an external IdP to issue an assertion that another AS will accept as a grant, and ends up holding a backend access token for the user. The flow runs server-to-server, but the consequence is that any mis-validation in the chain turns vMCP into a confused deputy. The subsections below cover the artefacts vMCP handles and the trust assumptions it relies on.
+
+### 4.1 ID Token Exposure and Redaction
+
+The new `Identity.UpstreamIDTokens` field carries the user's upstream ID token in process memory. ID tokens are sensitive — more so than access tokens, because they identify the user directly. The field is redacted in `Identity.MarshalJSON` via the `SafeIdentity` marshalling path, exactly the way `UpstreamTokens` is already redacted. Because every JSON serialisation of an `Identity` goes through `MarshalJSON`, the token value never reaches structured logs, the audit pipeline, or outbound webhooks. The strategy itself never logs the token; it parses the `sub` claim for logging only and discards the rest.
+
+### 4.2 ID-JAG as a Short-Lived Bearer
```
tgrunnagle, line 510: **[F2, 8/10] 🔴 HIGH: Confused deputy — no validation of returned ID-JAG claims**

§4.3 argues that a token minted for one resource AS will be rejected by another AS. True — but the RFC does not specify that vMCP should validate the returned ID-JAG's `aud` against `TargetAudience` and `resource` against `TargetResource` before proceeding to Step B. A misconfigured or malicious IdP returning an ID-JAG with the wrong audience goes undetected at the vMCP layer; the resource AS catches it, but there is no defense-in-depth at the point where vMCP holds the assertion.

Suggested addition to §3.6 step 4 (after obtaining the ID-JAG): decode the token (without signature verification — vMCP is not in the trust chain) and assert `aud == TargetAudience` && `resource == TargetResource`. Return a hard error on mismatch and add a corresponding unit test case.

---

**[F4, 8/10] SSRF risk from operator-configured token URLs not addressed**

`IDPTokenURL` and `TargetTokenURL` are operator-supplied strings that vMCP POSTs to with sensitive credential material in the request body (`subject_token` in Step A, `assertion` in Step B). An operator with write access to `MCPExternalAuthConfig` could target internal cluster endpoints. The security section (§4) does not address this surface, nor does it specify that HTTPS is enforced at admission.

Suggest adding to §4: (1) `IDPTokenURL` and `TargetTokenURL` must be HTTPS — enforce at CRD admission; (2) the HTTP clients must not follow redirects into RFC-1918 space; (3) note that RBAC on `MCPExternalAuthConfig` write access is a security boundary. Cross-reference how `token_exchange` handles the same surface.

tgrunnagle (commented):
## Code Review: RFC-0079 ID-JAG Cross-App Access

| # | Finding | Score | Severity |
|---|---------|-------|----------|
| F1 | `TokenTypeIDJAG` URN may be missing `+jwt` suffix | 9/10 | HIGH |
| F2 | Confused deputy: no validation of returned ID-JAG `aud`/`resource` | 8/10 | HIGH |
| F4 | SSRF risk from operator-configured token URLs | 8/10 | MEDIUM |
| F9 | `SubjectProviderName` defaulting undefined for multi-upstream deployments | 7/10 | MEDIUM |
| F16 | ID-JAG replay attack surface (`jti` tracking) not discussed | 7/10 | MEDIUM |

This RFC is well-researched and architecturally sound — PoC-validated, strictly additive, with an unusually thorough security analysis for this repo. Two HIGH-severity protocol and security issues need attention before merge.

**F1** (line 247): The `TokenTypeIDJAG` constant is `urn:ietf:params:oauth:token-type:id-jag`, but the JWT header `typ` is `oauth-id-jag+jwt`. By the IETF typed-JWT convention (cf. `at+JWT` for JWT access tokens), the URN likely needs the `+jwt` suffix: `urn:ietf:params:oauth:token-type:id-jag+jwt`. If the draft registers this as the canonical form, the Step A request sends a wrong `requested_token_type` to any conformant IdP and the `issued_token_type` check in §3.6 will never match — a concrete interoperability failure. Please verify against draft-ietf-oauth-identity-assertion-authz-grant-04 §3. The xaa.dev sandbox may have non-conformant behavior that masked the issue in the PoC.

**F2** (line 528): The strategy should decode the returned ID-JAG (without signature verification) and assert `aud == TargetAudience` and `resource == TargetResource` before proceeding to Step B. §4.3 correctly argues that a token minted for one AS is rejected by another, but vMCP holding no JWKS and not checking the returned claims means a misconfigured or malicious IdP response goes undetected at the vMCP layer. Defense-in-depth here is cheap: a single JWT parse with no crypto.

---
*Generated with [Claude Code](https://claude.com/claude-code)*

rfcs/THV-0079-cross-app-access-grant.md
```diff
@@ -0,0 +1,682 @@
...
+
+Both the ID-JAG and the backend access token are bearer credentials, and the design bounds their blast radius by lifetime and by never persisting them.
+
+- The **ID-JAG** is short-lived — 300 s typical in the sandbox — and held only in process memory for the duration of the `Authenticate` call; it is passed directly to Step B and not retained afterward. It is never written to disk, to a ConfigMap, to a RunConfig file, or to operator status.
+- The **backend access token** is cached per identity + backend in the existing `TokenCache`, which is in-memory with the access token's own `expires_in` as its TTL.
+
+Neither artefact outlives the process, and both are gated by short TTLs, so a memory-disclosure window is small and self-healing.
+
```
tgrunnagle, line 518: **[F16, 7/10] ID-JAG replay attack surface not discussed**

The ID-JAG structure (line 81) includes a `jti` claim whose purpose is replay prevention — but the security section does not mention this. Within the ~300 s validity window, a disclosed ID-JAG could be replayed at the resource AS if that AS does not track consumed `jti` values. vMCP's single-use, in-memory handling minimises the interception window, but it does not eliminate it.

Suggest adding a sentence to §4.2: "Replay prevention within the `exp` window is an explicit trust assumption on the resource AS implementing `jti` tracking; vMCP's own handling (no persistence, direct pass-through to Step B) minimises the interception surface but does not substitute for AS-side `jti` enforcement."

tgrunnagle (dismissed):
Overall LGTM. Automated review flagged a few issues to consider.

rfcs/THV-0079-cross-app-access-grant.md
jhrozek: Thanks for flagging this — worth double-checking carefully given how easy it is to conflate the two namespaces here.

I traced this against the actual draft text (revisions -00 through -04 — no later revision exists) plus the IANA OAuth Parameters registry, Okta's live cross-app-access implementation, and the `@modelcontextprotocol/typescript-sdk` PR #1328 code. All agree: the token-type URN used in `requested_token_type`/`issued_token_type` is `urn:ietf:params:oauth:token-type:id-jag` with **no** `+jwt` suffix (draft §4.3), and this has been consistent since draft -00.

The `+jwt` you're seeing belongs to a separate field — the JWT `typ` header (draft §3.1: `MUST be oauth-id-jag+jwt`), which follows RFC 8725's typed-JWT convention independently of the RFC 8693 token-type URN namespace. The `at+jwt` analogy doesn't transfer here because `at+jwt` (RFC 9068) is itself a `typ` header value, not a token-exchange URN precedent.

So the RFC's `TokenTypeIDJAG` constant is correct as written — no change needed. Good catch to verify though, since the two fields do look like they should match at a glance.

rfcs/THV-0079-cross-app-access-grant.md
jhrozek: Confirmed — this is a real gap, thanks for calling it out.

Traced it into the actual code: `InjectSubjectProviderNames` (`pkg/vmcp/config/defaults.go`) defaults `SubjectProviderName` via `authserver.ResolveFirstUpstreamName(names)`, which silently returns `names[0]` regardless of how many upstreams are configured. Each upstream is already required to have an explicit, distinct name once there are 2+ (`validateUpstreamName`), but nothing stops the strategy-level defaulting from picking one of them arbitrarily when the strategy itself doesn't say which.

I've updated §3.7 to specify that `xaa` hard-errors on this ambiguity (rather than silently defaulting) when multiple upstreams are configured and the field is omitted — pushed in c392d0f. Filed stacklok/toolhive#5697 to track the implementation, scoped to `xaa` only since it has no existing deployments to break. `token_exchange`/`aws_sts` keep today's silent-default behavior for now — changing those without a migration path would be a breaking change for existing deployments, so that's deferred to stacklok/toolhive#5687's broader defaulting-helper consolidation.

rfcs/THV-0079-cross-app-access-grant.md
jhrozek: Good catch on the gap — added to §4.2 (pushed in 0133010).

One correction on sourcing while verifying: I checked the actual draft text and "replay" doesn't appear anywhere in it, and the security-considerations text doesn't assign jti-tracking responsibility to either the Client or the Resource AS — the draft is simply silent, not affirmatively pointing at the Resource AS. So I wrote it as ToolHive's own trust assumption (consistent with RFC 7523's inherent bearer-assertion replay exposure) rather than attributing it to draft text:

"Replay prevention within the exp window is a trust assumption ToolHive makes on the resource AS implementing jti tracking, consistent with RFC 7523's inherent bearer-assertion replay exposure — the draft itself is silent on replay and does not assign this responsibility explicitly. vMCP's own handling (no persistence, direct pass-through to Step B) minimises the interception surface but does not substitute for AS-side jti enforcement."

rfcs/THV-0079-cross-app-access-grant.md
jhrozek: Agreed this is worth adding, with one nuance: unsigned aud/resource inspection can't stop a malicious/compromised IdP (an attacker controlling the response can fabricate matching unsigned claims just as easily) — its real value is as a fast-fail config-hygiene check, catching IdP misconfiguration before an opaque Step B failure. The actual trust boundary stays the resource AS's own signature/aud/resource validation at Step B.

Since `xaa.go` is already merged (#5684) with the two-step flow live, this is an implementation fix rather than an RFC-text change — filed as stacklok/toolhive#5696, scoped to extend the existing `parser.ParseUnverified` call in `performStepA` (already used there for the `typ` header check) to also assert `aud`/`resource` match before proceeding to Step B.

tgrunnagle (approved):

PR #80, opened 2026-07-03
RFC: Project-level skills lock file

samuv:
## Summary

Proposes a project-level lock file (`toolhive.lock.yaml`) that pins the name, version, source, resolved reference, and digest of every project-scoped skill install, bringing skills the reproducibility guarantees that `package-lock.json`, `Cargo.lock`, and `go.sum` provide elsewhere.

- **Problem:** `thv skill install --scope project` leaves no shareable, version-controlled pin, so teammates cloning a repo get whatever the catalog currently serves.
- **Lock file:** committed at the project root, client-agnostic, entries sorted for stable diffs; designed as the *general* ToolHive project lock (`skills:` today, room for `plugins:` later).
- **New commands:**
  - `thv skill sync` — restore the pinned skill set (drift reported distinctly; `--check` gives CI a `npm ci`-style gate; `--prune` removes unmanaged skills).
  - `thv skill upgrade` — re-resolve each entry's original source and rewrite the pin on digest change; immutable pins (OCI digests, full commit hashes) are never re-resolved.
- **Out of scope:** manifest/constraint resolution, per-client pinning, dependency graphs, user-scope installs.

POC implementation: https://github.com/stacklok/toolhive/pull/5715


rfcs/THV-0080-skills-lock-file.md
JAORMX, line 1: ## 🔴 Critical: No binding between `source` and `digest` — committed digest is an unauthenticated trust root

**CWE-345, CWE-347, OWASP LLM01+LLM03**

The lock treats `digest` as an integrity anchor, but the design **never verifies that the pinned digest corresponds to the declared `source`**. The entry tuple `{name, source, resolvedReference, digest}` is entirely hand-editable in the committed YAML, and the validator (§Input Validation) checks `ValidateSkillName(name)` + digest *format* — not *authenticity*. The claim that "a tampered lock can only pin references that would also pass a manual `thv skill install`" is a non-sequitur: a manual install pulls *whatever digest the registry currently serves for that reference*; a lock entry pulls *a specific digest the lock author wrote down*. Those are different checks.

**Attack:** an attacker with PR access (the exact threat this section names) writes:
```yaml
- name: code-review
  source: code-review
  resolvedReference: ghcr.io/org/code-review:1.0.0
  digest: sha256:<digest of attacker image pushed to a typosquat/different registry>
```
Name-validation passes, SSRF passes (not localhost), digest-format passes. Nothing compares the digest against "what the canonical source actually published" — there is **no GOSUMDB, no transparency log, no signature, no per-source allowlist.** The malicious AI-instructions image installs on every syncing teammate.

**Why this is worse than npm, not parity with go.sum/Cargo.lock:** `go.sum` is backed by **GOSUMDB** — a Trillian-backed Merkle transparency log operated by Google at `sum.golang.org` ([go.dev/ref/mod#checksum-database](https://go.dev/ref/mod#checksum-database)) — so a tampered go.sum hash can be checked against a global append-only log. `Cargo.lock` is backed by crates.io's curated, immutable, content-addressed registry. npm made *this exact mistake* historically (malleable `resolved` URLs, integrity not always enforced on all paths). **This RFC is closer to npm than to go.sum/Cargo.lock** because the lock has no external record to bind source→digest against. And skills are *instructions an AI follows* — higher blast radius than inert code.

**Minimum v1 fix:**
1. Bind source→digest at first-write: record the digest *as resolved by the tool from `source`*; at load, re-derive the expected digest from `source` (fresh authenticated resolution) and require the committed `digest` to match either (a) the currently-resolved digest or (b) a digest previously observed by a prior install/upgrade on this machine (local append-only "seen digests" ledger — the go first-write-wins analogue).
2. `sync` should refuse to install a lock entry whose pinned digest has never been observed from the declared `source` unless `--trust-lock` is passed.
3. Add a `registry:`/`catalog:` field recording the canonical origin, so a digest moved between registries is visible in the diff.
4. State explicitly in §Threat Model that a committed digest is an *assertion, not a verification*, until signatures land.

JAORMX, line 1: ## 🔴 Critical: Accepting SHA-1 git commit hashes as the integrity primitive for a new 2026 supply-chain system

**CWE-327, CWE-345, OWASP LLM03**

"Inherited from git" is not a defense for a *new* system shipping in 2026. SHAttered (2017) and the 2020 Leurent-Peyin chosen-prefix work (~2⁶³, ~$45k/GPU-year, cheaper yearly) made SHA-1 collision attacks practical. The relevant attack here is the cheaper **chosen-prefix** path: an attacker publishes a *pair* of repos — a benign one whose commit hash the victim pins, and a malicious one sharing that prefix — and the lock cannot tell them apart, because `resolvedReference: git://github.com/org/skills.git` + a 40-hex `digest` is the entire identity. Worse, combined with the source↔digest unbinding (see comment on §Input Validation), even a non-colliding SHA-1 can be swapped between repos.

**What the comparators actually do (verified):**
- **go.sum** uses **SHA-256 only** (`h1:`), hashing the module *zip* with SHA-256 — it never trusts a git commit hash as the integrity anchor ([go.dev/ref/mod#go-sum-files](https://go.dev/ref/mod#go-sum-files): "Currently, SHA-256 (`h1`) is the only supported hash algorithm").
- **Cargo.lock** records `"registry+...#<sha256-of-crate>"` — SHA-256 of the published tarball, never a bare SHA-1.
- **npm** uses SHA-512 `integrity` (legacy `sha1` now deprecated/ignored by default).

All three avoid SHA-1 for content integrity. **This RFC is worse than all three on the git path.**

Additionally, the git pin pins a **commit hash** (history), not a **tree hash** (content-at-path). For a multi-file skill (`SKILL.md` + `scripts/` + `references/`), the commit hash doesn't uniquely identify the content of the skill subdirectory. `go.sum` hashes the module zip contents; `poetry.lock` stores a SHA-256 *per file* (sdist + wheel). A git commit hash is weaker than both.

**Fix:**
1. **Require** sha256 git repos (`extensions.objectFormat=sha256`) for git-pinned skills; record a sha256 commit hash. Reject 40-hex SHA-1 git pins at load, or require an explicit `--insecure-sha1`.
2. For git pins, store the **git tree OID of the skill subdirectory** (content, not history), or a SHA-256 of the materialized file set.
3. Stop calling SHA-1 "inherited" — it's an *elected* weakness given git offers sha256 and no comparator uses SHA-1. Rewrite this bullet to acknowledge it as a known weakness with a migration path, not an accepted trade-off.

JAORMX, line 1: ## 🔴 Critical: Transitive skill dependencies are not locked — the canonical npm reproducibility mistake

A skill declaring `toolhive.requires: [other-skill]` materializes the dependency into SQLite with **no lock entry**. Two machines running `sync` against an identical `toolhive.lock.yaml` will: install the locked top-level skill at its pinned digest, then resolve `other-skill` from whatever the catalog *currently serves*. **That is non-reproducible by construction.**

This is the precise shape of the npm incident class the operator wants to avoid: a top-level pin gives false reproducibility while a transitive artifact floats. For skills (AI instructions), two teammates' agents operate under different instructions and nobody notices.

**Verified comparators:** npm `package-lock.json` v2/v3 locks the *entire* transitive graph inline; `Cargo.lock` stores one entry per transitive crate and locks the full graph ([doc.rust-lang.org/cargo/reference/resolver.html](https://doc.rust-lang.org/cargo/reference/resolver.html)). **Every mature ecosystem locks transitive deps.** The RFC explicitly does not (Non-Goals).

**Fix:** v1 should at minimum write lock entries for transitively-materialized skills (the SQLite `skill_dependencies` table already records them) with a `requiredBy:`/`provenance:` field. `sync` then installs them at their pinned digest. This matches Cargo.lock's model (record what was installed, don't build a resolver) without requiring a constraint resolver. The "install them explicitly at project scope" workaround doesn't scale and doesn't remove the unpinned-transitive risk for users who don't do it.

As a side effect, this also creates an odd tension with `--prune`: an unmanaged dependency is a prune candidate, so a habitual `sync --prune` could delete a skill that a locked skill `requires`. Recording transitive deps as lock entries (even `managed: false` with a provenance marker) would make them *not* prune candidates.

JAORMX, line 1: ## 🔴 Critical: `sync` compares against SQLite, not re-hashed content — the CI gate can pass on tampered installs

**CWE-345, OWASP API9**

The CI gate (`sync --check`, line 86) is advertised as analogous to `npm ci` / `cargo --locked`. But:
- **`npm ci` re-hashes installed files against the lock's `integrity` field** — that's the whole point of `ci`.
- **`go mod verify` re-hashes the module cache** ([go.dev/ref/mod#go-mod-verify](https://go.dev/ref/mod): "hashes each downloaded module `.zip` file and extracted directory, then compares those hashes with a hash recorded when the module was first downloaded").

The RFC's `--check` verifies against the **local SQLite DB**, which is mutable and which this paragraph admits can diverge from on-disk content. An attacker (or a buggy editor, or a malicious skill script that rewrites its own `SKILL.md` post-install) is **not detected** by the CI gate. For *prompt-instruction* artifacts, undetected on-disk tamper is high-impact: the AI follows tampered instructions while CI is green. The `upToDate` status is silently false here, with no operator-visible signal that files weren't re-hashed.

Additionally, the RFC doesn't specify that digest verification is **always-on on every install path including OCI cache hits**. The Yarn-classic lesson (no integrity enforcement by default) is that enforcement must be default, never opt-in.

**Fix:**
1. `sync --check` must **re-hash on-disk skill content** against the locked digest, not just read the SQLite record. This is the `go mod verify` / `npm ci` analogue.
2. Specify digest verification is **always-on on every install path including OCI cache hits**.
3. If re-hash is deferred, rename the guarantee: `--check` verifies install *records* match the lock, not supply-chain *integrity* — and stop calling it the `npm ci`/`cargo --locked` analog (line 86).
4. Per-client file-presence checks (also noted here as future work) should be part of `--check`, since an empty skill directory reporting `upToDate` defeats the gate.

**POC divergence (verified against toolhive#5715):** `sync.go:57-58` confirms `syncEntry` compares `existing.Digest == entry.Digest` against the SQLite `store.Get` result — no file re-hash. This matches the RFC's stated design, so the gap is in the design, not just the POC.

JAORMX, line 1: ## 🔴 Critical: `sync` is a one-command prompt-injection amplifier with no pre-install gate

**CWE-693, OWASP LLM01**

This section says sync "turns 'review what you install' into 'run one command after `git pull`', silently installing instructions an AI assistant will follow." The §Mitigations (line 148) then claims the flow is "observable rather than silent" because "sync reports new, drifted, and pruned entries per skill." But that report is **post-install**: the Sync design (line 84) is "install resolvedReference@digest" → report. There is no pre-install summary, no `[y/N]` prompt, no `--yes` flag anywhere in the spec. A new teammate running `thv skill sync` after `git pull` has skills installed into their client dirs before they see a single line about what was installed.

For a supply-chain artifact an AI will follow, post-hoc reporting is too late: the malicious skill is already installed and will be followed by the assistant before the user reads the report. The "observable" mitigation is a reporting control, not a gating control.

**Fix:**
1. `sync` should, by default (interactive TTY), print "About to install N skills (M new, K drifted) from:" with name/source/digest, then `Install? [y/N]` defaulting to **N** (stray Enter = no-op). `--yes` skips for scripted/CI use; non-interactive without `--yes` fails closed.
2. Track a local "last synced lock revision" in SQLite and require `--trust-new` for entries added since the last sync — narrows a malicious PR's blast radius to "first teammate who runs `--trust-new`" instead of distributed-silent.
3. Upgrade §Mitigations line 148 from "observable" to "gated" — observable-after-the-fact is not a supply-chain control for prompt-instruction artifacts.

JAORMX, line 1: ## 🟠 High: `source` re-resolution on upgrade is a catalog-redirect / typosquat vector (TOCTOU)

**CWE-345, OWASP LLM03**

When `source` is a bare registry name (e.g. `source: code-review`), `upgrade` re-resolves it "through the catalog." The RFC doesn't specify the catalog mapping is signed, pinned, or authenticated to a specific catalog instance. An attacker who can move a tag, typosquat (`code-review` vs `code-revew`), or influence which catalog the resolver consults can make `upgrade` re-resolve `source: code-review` to attacker content — which then gets a *fresh, legitimately-formatted* digest and is committed by the developer. The §Authentication and Authorization section says "reuse existing OCI Docker credential chain and git token auth paths" — that authenticates *transport* (TLS, registry bearer), not *mapping* (name→reference). This is a TOCTOU between resolve and install.

This is the same shape as npm's classic `resolved` URL malleability problem (arbitrary tarball URLs), which enabled the event-stream/ua-parser-js class of attacks.

**Fix:**
1. Pin the catalog/registry identity in the lock entry (`catalog:` field with a version/digest of the catalog index consulted at install time); refuse cross-catalog re-resolution without `--allow-source-change`.
2. After re-resolution on upgrade, require the new `resolvedReference` to either match the locked one (no-op) or be blessed by a signature/sumdb lookup. Do not install content whose reference changed without an external attestation.
3. Treat mutable `source` (registry name, git branch/tag) as higher-risk than immutable (full OCI digest, full 40/64-hex commit). Require `--allow-mutable-upgrade` or record the resolved-at time.
4. Recommend teams prefer fully-qualified OCI refs (`ghcr.io/org/code-review`) in `source` for security-sensitive skills to avoid catalog re-resolution entirely.

JAORMX, line 1: ## 🟠 High: `sync --check` ≠ `cargo --locked` / `npm ci` — the analogy oversells the guarantee

**Verified comparator semantics:**
- **`cargo build --locked`** ([doc.rust-lang.org/cargo/commands/cargo-build.html](https://doc.rust-lang.org/cargo/commands/cargo-build.html)): "Asserts that the exact same dependencies and versions are used as when the existing `Cargo.lock` file was originally generated. Cargo will exit with an error when ... Cargo attempted to change the lock file due to a different dependency resolution." → refuses lock **mutation** in CI.
- **`npm ci`** ([docs.npmjs.com/cli/v10/commands/npm-ci](https://docs.npmjs.com/cli/v10/commands/npm-ci)): clean install from lock, **never mutates the lock**, fails if lock is out of sync with `package.json`.
- **`go build -mod=readonly`** (default since Go 1.16): refuses implicit `go.mod`/`go.sum` mutation.

Since RFC-0080 has **no manifest** (Non-Goals), there's no constraint for the lock to be "in sync with." So `--check` gates *installed state vs. lock* (are the right skills installed), not *lock vs. manifest* (is the lock up to date). That's a different, weaker guarantee than `--locked`. `--locked`'s real value is catching "developer bumped the catalog but forgot to commit the lock update" — RFC-0080's `--check` doesn't address this, because install never *mutates* the lock (sync "never rewrites the lock").

**Fix:**
1. Clarify `--check`'s contract: it verifies *installed FS/DB state matches the lock*, full stop. Don't market it as the `--locked` analog.
2. To get the real `--locked` guarantee (lock is not stale), add: `thv skill upgrade --dry-run --check` (or `sync --verify-lock`) that re-resolves every `source` and fails if any would produce a different digest — i.e., the lock is stale. Without it, a stale lock is only caught when someone manually runs `upgrade`.
3. Ensure `--check` re-hashes content (see comment on line 85) — otherwise CI greenlights a SQLite-only check that's a tautology if both lock and DB are compromised together.

JAORMX, line 1: ## 🟡 Medium: `--dry-run` is a misnomer — it fetches to cache/clones (network + cache writes)

**CWE-400 (resource consumption), UX hazard**

The RFC honestly discloses that dry-run "still pulls the artifact into the local cache or clones the repo." But every other tool a developer reaches for (`cargo --dry-run`, `npm install --dry-run`, `git fetch --dry-run`, `terraform plan`) treats `--dry-run` as **no network, no writes, safe in a read-only CI sandbox.** Here it makes network calls **and writes to the local cache/clone.** The name violates the expectation of exactly the cautious operator who reaches for it.

This also has a security dimension: a CI job doing `upgrade --dry-run` to preview available updates will populate the cache, affecting subsequent jobs or masking "would this pull succeed fresh?" questions. And `--dry-run` that *fetches attacker-controlled bytes* into the cache (even without extracting) is a larger surface than users expect — DoS via large blobs, slowloris registries, git clone bombs.

**Fix:**
1. Rename to `--preview` (or `--plan`), or keep `--dry-run` but mandate that `--help` and the command's startup banner print, verbatim: *"NOTE: --dry-run still fetches artifacts into the local cache; it skips only filesystem and database writes, not network or cache writes."*
2. Bound dry-run fetches (max body size, `context.WithTimeout`).
3. For the CI gate use case (`--check`), ensure there's a path that does *not* populate the cache, so CI gates are hermetic.
4. Consider a true `--no-network` mode that only reports based on cached state.

JAORMX, line 1: ## 🟡 Medium: `SkillService` Go interface break is a real, breaking change — "if any exist" is a hedge against a verified yes

**Verified:** `SkillService` is an **exported** interface in `github.com/stacklok/toolhive/pkg/skills` (`pkg/skills/service.go`), with `//go:generate mockgen`. It has concrete implementers with compile-time assertions (`var _ skills.SkillService = (*Client)(nil)`) in at least two other Stacklok repos:
- `toolhive-git-skills/pkg/skills/client/client.go`
- `stacklok-enterprise-platform/enterprise/toolhive/upstream/pkg/skills/client/client.go`

Adding `Sync`/`Upgrade` to the interface will cause those `Client` implementations to fail compilation on their next rebase. The "if any exist" hedge should be dropped — the answer is yes.

This also contradicts §Backward Compatibility (line 177: "This change is fully backward compatible"), creating an internal inconsistency in the RFC.

**Fix:**
1. Segregate `Sync`/`Upgrade` into a separate `SkillLocker` (or `LockSyncer`) interface that `*skillsvc.service` satisfies additionally. Lock-file operations are conceptually distinct from skill CRUD; the split is defensible on its own merits and avoids breaking existing `SkillService` implementers.
2. If they must live on `SkillService`, commit to a documented `SkillServiceV2` and keep v1 intact, or ship a default-implementation embed (`SkillServiceDefault`) implementers can embed to get no-op `Sync`/`Upgrade`.
3. Reconcile §API Changes and §Compatibility: state explicitly that the *HTTP* change is additive but the *Go interface* change is breaking, and drop "fully backward compatible."
4. If external implementers must honor the `source`-immutability contract (line 91: `installInternal` not `Install`), that contract must be **public** — `installInternal` is unexported, so external implementers have no access to the "don't clobber `source`" path and will silently overwrite it.

JAORMX, line 1: ## 🟡 Medium: Best-effort lock write can cause silent stale lock and silent downgrade on sync

**CWE-345**

"Lock write errors never fail the install" means a successful install with a failed lock write leaves the committed lock *silently stale* for any teammate who pulls. The "warning" surfaces to the *current* user, but if they ignore it (or it's a CI install that doesn't fail the build), the staleness propagates. The next `sync` by a teammate *overwrites* their correct install with the stale lock's (older) digest — i.e., a **silent downgrade** — because sync "reinstalls at the pinned `resolvedReference@digest`."

**Comparator:** `Cargo.lock` write failure **fails the build** — the lock is atomic with install. `go mod` commands fail on lock write inconsistency. The RFC's "warn and continue" is strictly weaker than both comparators it invokes.

**Fix:**
1. For `install --scope project`, the lock write should be **part of the install transaction** — fail the install (non-zero exit) if the lock can't be written, with a clear "lock not updated; do not commit" message. "Warn and continue" is acceptable for *user-scope* (excluded from the lock), but not project-scope where the lock is the contract.
2. At minimum, `sync` should *refuse to downgrade* an entry whose installed digest is newer (by catalog version or recorded install timestamp) than the lock's pinned digest, requiring `--force` — this prevents the silent-revert footgun.
3. A failed project-scope lock write must exit non-zero so CI catches it, not just warn.

JAORMX, line 1: ## 🟡 Medium: `projectRoot` validation is underspecified at the trust boundary

**CWE-22 (Path Traversal), CWE-345**

"Validated the same way existing project-scope install paths are" is vague for *the* new trust boundary this RFC introduces: an API body field `projectRoot` that selects where a lockfile is read and written. The RFC doesn't specify:
- (a) canonicalization + containment check (reject `..`, reject symlinks escaping an allowed root)
- (b) whether arbitrary absolute paths are accepted (can an API caller point `projectRoot: /etc` and have the daemon write `/etc/toolhive.lock.yaml`?)
- (c) whether the daemon must be co-located with the project (a remote `thv serve` daemon serving a `POST /skills/sync` with a caller-chosen `projectRoot` is a different threat than a local CLI)

**Fix:** Specify concretely in §Input Validation: `projectRoot` must (1) be absolute, (2) be canonicalized with symlinks resolved, (3) contain a `.git` directory or be within a `.git`-rooted tree (matching the auto-detection rule), (4) reject if it resolves outside the invoking user's home or a configured allow-root. For the HTTP API specifically, require `projectRoot` to be under a daemon-configured `--serve-root` and reject otherwise — an API caller must not be able to point the daemon at arbitrary filesystem locations. Cite CWE-22 explicitly.

JAORMX, line 1: ## 🟡 Medium: `--prune` conflates "never-managed" with "removed-from-lock-by-tampered-PR"

**CWE-693**

This section names the threat ("a tampered lock that *removes* entries, combined with a habitual `sync --prune`, deletes legitimate skills") and the mitigation is "opt-in" + "itemized." But:
1. **No pre-delete prompt.** The itemization is *post-deletion*. A destructive flag whose only disclosure is a post-hoc report means a `sync --prune` after a tampered PR deletes legitimate skills before the operator can read the line.
2. **"unmanaged" is overloaded.** It simultaneously means (a) "installed locally but never in the lock" and (b) "was in the lock at last sync, removed from lock since." A tampered PR removing entries produces case (b), which is far more dangerous. Both collapse to `unmanaged` in the report. A security-policy skill intentionally installed out-of-band by an org SecOps team (never committed to the repo lock) would also be pruned under the current design.

**Fix:**
1. Distinguish `pruned: removed-from-lock` from `pruned: never-managed` in the report, and require **separate confirmation** (or separate flags `--prune-removed` vs `--prune-unmanaged`) for the two.
2. Add a `managed` marker in the SQLite record set at install time ("this install was project-scope-locked" vs "this install was out-of-band"), so `--prune` only removes skills the lock *used to* manage — never skills installed by other means.
3. `--prune` should default to a prompt listing each itemized deletion before acting; `--prune --yes` for scripted use.

JAORMX, line 1: ## 🟡 Medium: `failed` collapses unrecoverable / retryable / policy-rejected causes — a caller cannot branch

The RFC enumerates failure causes — "registry down" (retry), "digest garbage-collected" (re-pin via upgrade, do not retry), "SSRF blocked" (fix policy or the lock, do not retry) — plus lock-write-failure-after-upgrade. But `SyncResult.Failed` / `UpgradeOutcome` carry only a human-readable `Error` string. A CI caller that gets `failed` cannot decide retry vs. fail-the-pipeline vs. page-a-human; it must string-match `Error`. This is the surface every CI consumer gates on.

**Fix:** Give `SyncFailure`/`UpgradeOutcome` a typed `Reason` enum (`registry-unreachable`, `digest-missing`, `validation-rejected`, `lock-write-failed`, `unknown`) alongside the free-text `Error`, or return typed errors from `Sync`/`Upgrade`'s per-entry path. At minimum commit to a sentinel set for the three causes the RFC already names.

Also: exit-code semantics are undocumented. Three semantically different non-zero outcomes share one signal (drift detected / registry down / validation rejected). A CI pipeline author cannot write `if exit != 0` differently for these. Define a small exit-code table: `0` clean, `2` drift detected (`--check` only), `3` partial failure (see report), `4` validation/policy rejection.

JAORMX, line 1: ## 🟡 Medium (architecture): General project lock defined under a skills RFC — layering violation

**Spec adherence to THV-0077:** This RFC declares "`toolhive.lock.yaml` is the general ToolHive project lock file, not a skills-only artifact" and reserves the `plugins:` namespace for THV-0077. But **THV-0077 has zero mention of `toolhive.lock.yaml` or a `plugins:` lock key** (verified by grep — its only "locking" is SQLite row-locking at the per-plugin level). RFC-0080 silently commits THV-0077 to a contract THV-0077 never specified.

Defining the cross-artifact top-level schema (version field, sibling-key extension contract, "hard-error on unknown version" policy that binds all future artifact types) belongs in a dedicated project-lock RFC, or must be co-authored with THV-0077, not asserted in a skills RFC. A future plugins RFC author shouldn't discover the lock-file contract is already locked in by a skills RFC.

**Fix:** Either (a) narrow this RFC's scope to `toolhive.skills.lock.yaml` (skills-only) and defer the general lock to a separate RFC, or (b) if the general lock name is kept, explicitly state that the `plugins:` schema is deferred to THV-0077 and this RFC only owns the `version` + `skills:` keys, with the sibling-key extension being a *contract proposal* that THV-0077 must ratify.

## 🟡 Medium (spec): THV-0041 duality unacknowledged

THV-0041's goal is "SQLite as the *unified* local state store," explicitly to kill "bespoke file-locking schemes." This RFC reintroduces a committed YAML as a second source of truth with `sync` to reconcile. Not a schema contradiction (the lock is committed, SQLite is local — different concerns), but "cleanly additive" is overstated: it creates lock-vs-DB drift (which the RFC itself admits sync cannot fully detect, line 85). The RFC should explicitly reconcile why a committed second store does not violate 0041's single-source-of-truth goal, rather than claiming "No SQLite schema changes are required" (technically true but evades the architectural point).

JAORMX, line 1: ## 🟡 Medium: Lockfile discovery — "nearest `.git` directory" breaks on worktrees and submodules

"Auto-detecting the git root from the current working directory, using the nearest enclosing `.git` directory." In a git **worktree**, `.git` is a *file* (containing `gitdir: ...`), not a directory. "Nearest enclosing `.git` directory" will misbehave on worktrees and submodules — exactly the layouts large teams use.

**Fix:** Say "nearest `.git` (file or directory), resolving a worktree pointer to its common dir" — or, more honestly, use `git rev-parse --show-toplevel` / `--git-common-dir` semantics. Separately, the no-`.git` case "errors rather than guessing" — good, but the error string must be `"no git repository found (or not inside one); pass --project-root <dir> to specify the project root"` (cause + next action), not just "no git repository."

## 🟡 Medium (UX): First-run adoption — no `--adopt`, existing installs threatened by `--prune`

The team that commits `toolhive.lock.yaml` for the first time, on a machine where skills are already installed, runs `sync` and sees every existing skill flagged `unmanaged` — the same flag that `--prune` deletes. An operator who reads "unmanaged" + knows `--prune` exists will reasonably fear their legit installs. There is no `thv skill sync --adopt` (or `--reconcile`) to claim existing installs into the lock.

**Fix:** Specify an adoption path, or at minimum a distinct first-run message: *"No lock entries yet, but N skills are installed locally. They are unmanaged. To pin them, run `thv skill sync --adopt` (writes current digests to the lock) or `thv skill install --scope project` each."*

JAORMX, line 1: ## 🟢 Low: `installedAt` omission is well-reasoned — matches all comparators

Verified: `go.sum`, `Cargo.lock`, `poetry.lock`, `package-lock.json` all store identity + source + integrity, not chronology. The rationale here (timestamps produce merge churn; `git blame` answers "when pinned") is correct and matches ecosystem consensus. No action — just confirming this design decision aligns with prior art.

One minor suggestion: add a one-line `--help` pointer ("Skill pin history: `git log -p -- toolhive.lock.yaml`") so operators don't ask "where's the timestamp?"

JAORMX (commented):
## Panel review — cross-cutting summary

Conducted a three-axis panel review (spec / standards / domain) with specialist reviewers (secure-code-reviewer, software-architect, library-reuse-reviewer, devex-reviewer, ux-reviewer) plus verified research against go.sum, Cargo.lock, npm, pnpm, and poetry.lock. All factual claims were verified against source code/docs before posting. Inline comments below cover location-specific findings; this comment captures the cross-cutting picture and the non-location-specific ones.

### Overall

The RFC is well-researched and consciously cites the right prior art. It correctly avoids several npm mistakes (pins content digests not registry names, refuses to rewrite `source`, treats `--prune` as opt-in, hard-errors on unknown schema versions). But it **inherits npm's exact supply-chain gaps while claiming parity with `go.sum`/`Cargo.lock`** — and go.sum/Cargo.lock each close those gaps with an *external trust root* (GOSUMDB / curated registry) that this design lacks. For skills (instructions an AI executes), the stakes are higher than inert code.

### Critical findings (inline comments posted)

1. **No binding between `source` and `digest`** — a committed digest is an assertion, not a verification. A tampered PR can pin any digest that passes format/name validation, and `sync` faithfully installs malicious AI instructions. go.sum has GOSUMDB (Trillian Merkle transparency log, Google-operated); Cargo.lock has crates.io's curated content-addressed registry. This RFC has neither.

2. **SHA-1 git commit hashes accepted for a new 2026 supply-chain system** — all three comparators (go.sum SHA-256, Cargo SHA-256, npm SHA-512) avoid SHA-1 for content integrity. The "inherited from git" defense doesn't hold for a new system when git offers sha256 and no comparator uses SHA-1.

3. **Transitive skill dependencies (`toolhive.requires`) not locked** — two machines with identical locks can install different transitive skill content. Every mature ecosystem (npm v2/v3, Cargo.lock) locks the full transitive graph. This is the canonical npm reproducibility mistake.

4. **`sync`/`--check` compares against SQLite, not re-hashed content** — `npm ci` re-hashes files against the lock; `go mod verify` re-hashes the module cache. This RFC's CI gate can pass on tampered installs. For prompt-instruction artifacts, undetected on-disk tamper is high-impact.

5. **`sync` is a one-command prompt-injection amplifier with no pre-install gate** — the RFC's own threat model names this, but the mitigation is post-install *reporting*, not pre-install *gating*. For AI-instruction artifacts, reporting after install is too late.

### High findings (inline comments posted)

- **`source` re-resolution on upgrade is a catalog-redirect/typosquat vector (TOCTOU)** — same shape as npm's `resolved` URL malleability.
- **No signature/transparency log/provenance in v1** — "future Sigstore" is a forward pointer, not a v1 control. THV-0030:472 commits to Sigstore; this RFC never reconciles.
- **`sync --check` ≠ `cargo --locked`** — the analogy oversells the guarantee. There's no command that catches a stale lock (dev bumped catalog, forgot to commit lock update).

### Medium findings (inline comments posted)

- **`SkillService` Go interface break is real** — verified: exported interface with implementers in `toolhive-git-skills` and `stacklok-enterprise-platform`. "If any exist" should be "yes." §Compatibility's "fully backward compatible" contradicts §API Changes.
- **Best-effort lock write → silent stale lock / silent downgrade on sync** — Cargo makes lock write atomic with install (fails the build). This RFC's "warn and continue" is strictly weaker.
- **`--dry-run` is a misnomer** — fetches to cache/clones. cargo/npm/git `--dry-run` are side-effect-free.
- **`--prune` conflates "never-managed" with "removed-from-lock-by-tampered-PR"** — both collapse to `unmanaged`; a habitual `sync --prune` deletes legitimate skills indistinguishably.
- **`projectRoot` validation underspecified at the trust boundary** (CWE-22) — "validated the same way existing paths are" is vague for the one new API-body trust boundary.
- **`failed` collapses unrecoverable/retryable/policy-rejected causes** — callers can't branch; exit codes undocumented.
- **General project lock defined under a skills RFC** — THV-0077 has zero mention of `toolhive.lock.yaml`; this RFC silently commits it to a contract. Layering violation.
- **Lockfile discovery breaks on worktrees** — `.git` is a file, not a dir, in worktrees. Use `git rev-parse` semantics.
- **No `--adopt` path** — first-run treats existing installs as `unmanaged` (the same flag `--prune` deletes).

### The top three highest-leverage fixes (per the operator's stated priority: avoid npm, adopt Go/Rust rigor)

1. **Lock transitive dependencies** (write lock entries for materialized `toolhive.requires` skills with a `requiredBy` provenance marker — Cargo.lock records, doesn't resolve).
2. **Make `sync` re-hash content + always-on digest verification on every install path including cache hits** (`go mod verify` / `npm ci` rigor).
3. **Add a `publisher`/provenance field with change-detection on `upgrade`** — a cheap, local mitigation for the event-stream/ua-parser-js class ("digest correct, content malicious") that needs no Sigstore in v1, while explicitly naming GOSUMDB-style transparency logging + Sigstore (THV-0030) as the v2 trust layer.

### Verified comparators (sources cited in inline comments)

| Mechanism | Go | Cargo | npm | RFC-0080 |
|---|---|---|---|---|
| Hash algorithm | SHA-256 (`h1:`) | SHA-256 (tarball) | SHA-512 (SRI) | SHA-256 (OCI) **or SHA-1 (git)** ❌ |
| Transitive deps locked | ✅ | ✅ | ✅ (v2/v3) | ❌ |
| Re-hash on verify | ✅ (`go mod verify`) | ✅ (registry check) | ✅ (`npm ci`) | ❌ (SQLite only) |
| External trust root | ✅ GOSUMDB (Trillian) | ✅ curated registry | ⚠️ opt-in provenance | ❌ |
| Lock-write atomic | ✅ | ✅ (fails build) | ✅ | ❌ (best-effort warn) |
| CI frozen gate | ✅ `-mod=readonly` | ✅ `--locked`/`--frozen` | ✅ `npm ci` | ⚠️ `--check` (weaker) |
| Executable code on install | N/A | N/A | ⚠️ postinstall runs by default | ❌ undefined for skill scripts |

Happy to drill into any specific finding or help draft the `source`→`digest` binding design.

rfcs/THV-0080-skills-lock-file.md
samuv: Thanks @JAORMX — agreed that v1 cannot claim go.sum/Cargo.lock *parity* on trust.

**RFC changes (§Summary, §Trust model, §Alternatives 4):**
- Reframed honestly: in v1 a committed digest is an **assertion** verified against itself at install/sync, not against an external record. Trust root = PR review of the lock diff.
- Named v2 milestone: Sigstore (THV-0030) + transparency-log-style source→digest binding.
- Added Alternative 4 explaining why re-deriving digests from `source` at sync was rejected (defeats pinning).

We are **not** adopting the local "seen digests" ledger or `--trust-lock` gate — re-resolution at sync would make two machines diverge whenever the catalog moves, which is the reproducibility failure mode we're fixing.

samuv: Addressed with a new **`contentDigest: sha256:...`** field on every entry (§Detailed Design schema).

- `digest` remains the checkout pin (OCI manifest / git commit).
- `contentDigest` is a deterministic dirhash of the materialized skill file set — the content-integrity primitive for `--check` and always-on verification.
- Closes both the SHA-1 weakness and the "commit pins history, not skill subdirectory content" gap without requiring sha256 object-format repos (which most hosts don't offer yet).

Removed the "inherited from git" framing; git commit is now explicitly the fetch pin, not the integrity anchor.

samuv: Adopted the Cargo record-not-resolve model (§Transitive dependencies).

- Project-scope installs write lock entries for materialized `toolhive.requires` skills with `requiredBy: [parent]`.
- `sync` installs them at their pin; they are never `--prune` candidates while a parent requires them.
- Removed "no transitive locking" from Non-Goals (constraint *resolution* stays a non-goal).

This also fixes the `--prune` footgun where an unmanaged dependency of a locked skill could be deleted.

samuv: Fixed (§Component Changes, §Goals, §Testing).

- `sync --check` now **re-hashes on-disk skill content against `contentDigest`** in every detected client dir, plus file-presence checks — not SQLite records alone.
- Digest verification specified as **always-on on every install path**, including OCI cache hits.
- Dropped the `npm ci` / `cargo --locked` analogy; `--check` contract is stated plainly: installed state matches the lock.

samuv: Adopted (§Component Changes pre-install gate, §Mitigations).

- Interactive TTY: sync prints pre-flight summary (name/source/digest/contentDigest for install/drift/prune targets) and asks `Install? [y/N]` defaulting to **No**.
- `--yes` for scripts; non-interactive without `--yes` fails closed.
- Mitigations wording changed from "observable" to "gated".

samuv: Partial v1 fix (§Upgrade, §Threat Model).

- If re-resolution yields a **different `resolvedReference`** (not just a new digest for same ref), upgrade refuses without `--allow-ref-change` and highlights it in the report.
- Recommend fully-qualified OCI refs in `source` for sensitive skills.
- Catalog-identity pinning (`catalog:` field) deferred to v2 trust layer alongside Sigstore — same milestone as source→digest external binding.

samuv: Clarified (§Component Changes `--check`, §Upgrade).

- `--check` contract: verifies installed FS content matches lock (integrity gate). Not marketed as `--locked` analog.
- Added optional CI **freshness** gate: `thv skill upgrade --preview --fail-on-changes` exits non-zero if any mutable source would re-resolve to a new digest.
- `--check` now re-hashes content (see comment on SQLite-only check).

samuv: Renamed to **`--preview`** (§Component Changes).

- `--help` and startup banner state explicitly: "still fetches artifacts into local cache; skips only FS/DB writes."
- Bounded fetches (max body size, context timeout).
- `--check` does not populate cache (hermetic CI gate).

samuv: Fixed (§API Changes, §Compatibility).

- **`SkillLockService`** is a new interface with `Sync`/`Upgrade`; existing **`SkillService` unchanged** — compile compatibility preserved for `toolhive-git-skills` and enterprise platform implementers.
- Dropped "fully backward compatible"; HTTP additive, Go interface split is explicit.
- Source-immutability contract is now public via **`PreserveSource`** install option (replaces unexported `installInternal` reference).

samuv: Reversed earlier decision (§Component Changes).

- Failed lock write on **project-scope install now exits non-zero** with "skill installed but lock NOT updated — do not commit."
- Files/DB stay (no rollback), but command fails so CI/humans can't miss stale lock.
- §Compatibility notes this is a behavior change from POC warn-only semantics.

samuv: Specified concretely (§Input Validation, CWE-22).

- `projectRoot` must be absolute, symlink-canonicalized, within a git-rooted tree.
- HTTP API: constrained under daemon-configured `--serve-root`; cannot point daemon at arbitrary paths.

samuv: Hardened (§Component Changes, §Threat Model).

- SQLite **`managed`** marker set at project-scope locked install.
- `--prune` removes only **`removed-from-lock`** (previously lock-managed, no longer in lock); **`never-managed`** (out-of-band installs) reported but never pruned.
- Prune deletions itemized in pre-flight prompt; `--yes` for scripts.

samuv: Added (§API Changes).

- Typed **`reason`** enum on per-entry failures: `registry-unreachable`, `digest-missing`, `validation-rejected`, `lock-write-failed`, `ref-change-blocked`, `unknown`.
- Documented exit codes: `0` clean, `2` drift/check failure, `3` partial failure, `4` validation/policy rejection.

samuv: Addressed both points (§Detailed Design, §Forward Compatibility, §Data Model).

**Layering:** kept `toolhive.lock.yaml` name but reframed — this RFC **owns only `version` + `skills:`**; sibling keys (`plugins:`) are a *contract proposal* for THV-0077/future RFC to ratify, not an assertion.

**THV-0041:** added reconciliation in §Data Model — SQLite remains unified *local runtime* store; lock is shareable *committed pin*; sync reconciles the two.

samuv: Fixed (§Configuration Changes, §Component Changes `--adopt`).

- Discovery uses **`git rev-parse --show-toplevel` semantics** (worktree `.git` file handled).
- Error message includes cause + next action: pass `--project-root`.
- Added **`sync --adopt`** plus first-run message for existing installs without lock entries.

samuv: Acknowledged — no change needed to the omission itself.

Added the suggested **`--help` pointer** to `git log -p -- toolhive.lock.yaml` for operators asking "where's the timestamp?" (§Detailed Design).

rfcs/THV-0080-skills-lock-file.md
```diff
@@ -0,0 +1,310 @@
...
+- **Status**: Draft
+- **Author(s)**: Samuele Verzi (@samuv)
+- **Created**: 2026-07-03
+- **Last Updated**: 2026-07-07
+- **Target Repository**: toolhive
+- **Related Issues**: [toolhive#5715](https://github.com/stacklok/toolhive/pull/5715)
+
+## Summary
+
+Add a project-level lock file (`toolhive.lock.yaml`) that pins the name, version, source, resolved reference, checkout digest, and content digest of every project-scoped skill install (including transitively materialized dependencies), plus `thv skill sync` and `thv skill upgrade` commands to restore pinned state and pull newer content from the catalog. The lock borrows the *shape* of `package-lock.json`, `Cargo.lock`, and `go.sum` (identity, source, integrity in a committed file) but does **not** claim their external trust roots in v1: the trust root for a committed pin is PR review of the lock diff, with a v2 milestone for Sigstore signing (per THV-0030) and transparency-log-style source-to-digest binding.
```
JAORMX: I know you're explicitly calling out a v1 and v2, but can we just do v1? I'd really want to ship this with signing to begin with.

samuv: I was trying to defer this to a future update because I don't know much about Sigstore, but I'll try to do a full deep dive on it

samuv: @JAORMX Yes — pushed `cd35106` folding Sigstore into v1. The v2 trust-layer milestone is gone.

**What ships in v1:**

- **Sign on publish:** `thv skill push` keyless Sigstore signing (Fulcio + Rekor, OCI referrer); `--key` for key-based setups. Catalog skills signed in CI (coordinated workstream).
- **Verify on consume:** install/sync/upgrade verify signatures by default. Unsigned artifacts rejected unless `--allow-unsigned` (recorded as `unsigned: true` in the lock, visible in diffs).
- **Identity pinning:** each lock entry carries a `provenance` block (`signerIdentity`, `certIssuer`, optional `repositoryUri`/`sigstoreUrl`) — mirrors the existing registry provenance shape. Catalog-first on first install, TOFU for direct refs. Upgrade refuses signer changes without `--allow-signer-change`.
- **Git sources:** gitsign verification on pinned commits (unsigned git still needs `--allow-unsigned`).
- **Offline CI:** Sigstore bundles stored with install state; `sync --check` re-hashes `contentDigest` **and** re-verifies bundle + locked identity offline (hermetic, no Rekor round-trip).
- **Implementation:** extends toolhive-core `container/verifier` (sigstore-go, same stack as MCP image verification).

This directly closes the unauthenticated-trust-root finding via signature + pinned publisher identity + Rekor transparency, rather than the earlier "PR review only until v2" framing. See §Trust model, §Signing on publish, §Verification on consume, and Phase 5 in the Implementation Plan.

Happy to walk through any of the enforcement edge cases (`--allow-unsigned`, catalog-first vs TOFU, gitsign scope).

JAORMX (approved):

PR #83, opened 2026-07-21
Add RFC-0083: Stateless vMCP Support (MCP 2026-07-28)

jhrozek:
## Summary

Adds **RFC-0083: Stateless vMCP Support (MCP 2026-07-28)** — the decision-level counterpart to the vMCP stateless design work (toolhive#5756), and the vMCP sibling of [RFC-0081](rfcs/THV-0081-stateless-transport-proxies.md) (transport proxies), which explicitly defers the era-mismatched bridge cells here.

The 2026-07-28 revision removes protocol sessions, the `initialize` handshake, and all server-initiated requests. This RFC proposes making the Virtual MCP Server serve both the current (2025-11-25, "Legacy") and stateless (2026-07-28, "Modern") revisions on the same endpoint, discriminating **per request**, and — because vMCP terminates and re-originates on both edges — acting as the cross-generation bridge so a client on either revision can reach backends on either revision.

## Load-bearing decisions

- **Bypass, don't rebuild.** The vMCP core is already stateless; the Modern path invokes none of the session machinery (two-phase `SessionIdManager`, `MultiSession`, the session cache, Redis metadata, identity binding, cross-pod rehydration). Legacy is untouched.
- **Hand-rolled Modern dispatcher at a known seam, not gated on go-sdk v1.7.** Most of it is assembly of existing parts (classifier, the standalone `authzCallGate()` closure, `pkg/vmcp/conversion`); the one new piece — the JSON-RPC `resultType` envelope — is framed as a *temporary bridge* to be retired once v1.7's stateless server mode ships.
- **Per-backend dual protocol** with probe-Modern-first era detection cached in the health subsystem, and re-classification on backend-edge rejection as the correctness guarantee.
- **vMCP owns the off-diagonal bridge** (both era-mismatched cells), with synthesized Legacy-backend sessions partitioned by `(principal, backend)` and initialized with the calling principal's existing outgoing auth.
- **Server-minted, owner-bound workflow handles** — designed now (plaintext `(iss, sub)` reusing the existing `pkg/vmcp/session/binding` encoding, per #5306), enforcement sequenced with MRTR since the MVP exposes no client-visible handle.

## Notes

- Every codebase claim the RFC rests on was cross-checked against `toolhive` `origin/main`.
- No user-facing or CRD API changes for the MVP; fully backward compatible (all traffic classifies Legacy unless a request carries a Modern signal).
- Open Questions: none — all prior questions are resolved in the draft.

### DIFF
--- a/rfcs/THV-0083-stateless-vmcp.md
+++ b/rfcs/THV-0083-stateless-vmcp.md
@@ -1,0 +1,423 @@
+# RFC-0083: Stateless vMCP Support (MCP 2026-07-28)
+
+- **Status**: Draft
+- **Author(s)**: Jakub Hrozek (@jhrozek)
+- **Created**: 2026-07-21
+- **Last Updated**: 2026-07-21
+- **Target Repository**: toolhive
+- **Related Issues**: [toolhive#5756](https://github.com/stacklok/toolhive/issues/5756) (this),
+  [toolhive#5743](https://github.com/stacklok/toolhive/issues/5743) (epic)
+
+## Summary
+
+The MCP 2026-07-28 revision removes protocol sessions, the `initialize` handshake, and all
+server-initiated requests, replacing them with per-request `_meta`, a `server/discover` RPC, and
+Multi Round-Trip Requests (MRTR). This RFC proposes making the Virtual MCP Server (vMCP) serve both
+the current (2025-11-25, "Legacy") and the stateless (2026-07-28, "Modern") revision on the same
+endpoint, discriminating **per request** — and, because vMCP is an aggregating gateway that
+terminates the client connection and re-originates backend calls, acting as the **cross-generation
+bridge** so a client on either revision can reach backends on either revision.
+
+This is the decision-level counterpart to the vMCP stateless design work tracked in toolhive#5756,
+which carries the implementation detail. It
+is the vMCP sibling of [RFC-0081](THV-0081-stateless-transport-proxies.md) (transport proxies,
+toolhive#5755), which explicitly **defers the era-mismatched bridge cells to vMCP**.
+
+## Problem Statement
+
+The 2026-07-28 revision removes the mechanisms vMCP's session layer is built around: the
+`initialize` handshake, protocol sessions and `Mcp-Session-Id`, the standalone GET SSE stream and
+resumability, server-initiated requests (→ MRTR), and `ping`/`logging/setLevel`. The ecosystem will
+straddle both revisions for a long time — SDKs serve both on one endpoint and peers upgrade
+independently — so vMCP cannot do a flag-day cutover.
+
+Crucially, vMCP's stateful machinery — the two-phase `SessionIdManager`, per-session `MultiSession`
+objects, the 1,000-entry per-pod session cache, Redis session metadata, identity binding, and
+cross-pod tool rehydration — exists for **one reason**: to bridge vMCP's already-stateless core to a
+*session-ful protocol*. `coreVMCP` re-derives its advertised, per-principal view on every
+`List`/`Lookup`/`Call` from `(identity, backend registry, health)` and holds no MCP-session state.
+When the protocol stops being session-ful, that bridge is no longer needed on the Modern path — the
+design is mostly about **bypassing it on the Modern path**, not rebuilding it.
+
+RFC-0081 established per-request revision classification for the transport proxies and, in its
+compatibility matrix, marked the era-mismatched cells "Fails" for a pass-through proxy and deferred
+them to vMCP. vMCP is the component that can bridge them precisely because it terminates and
+re-originates on both edges.
+
+## Goals
+
+- Serve Legacy and Modern client traffic on the same vMCP endpoint, chosen per request.
+- Handle each backend according to its own negotiated revision (per-backend dual protocol).
+- Bridge all four client×backend era combinations, including the two RFC-0081 defers here.
+- Zero behavior change for existing 2025-11-25 traffic.
+- Add the minimum new machinery: reuse the existing `pkg/mcp` classifier; bypass the session bridge
+  on the Modern path rather than reimplement it.
+- Do **not** hard-gate the work on the go-sdk v1.7 bump.
+- Preserve every security invariant (auth per request, admission, rate limiting, audit) across the
+  Modern path, and relocate — not drop — the cross-request binding invariant onto workflow handles.
+
+## Non-Goals
+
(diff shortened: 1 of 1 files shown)


JAORMX (commented):
I pulled this RFC, the companion design doc (`toolhive@worktree-vmcp-stateless-design`), verified every load-bearing codebase claim against `toolhive@main`, and cross-checked every protocol claim against the authoritative MCP 2026-07-28 **draft spec** (changelog, lifecycle, `server/discover`, MRTR, caching, and the raw `schema.json`).

**Bottom line: well-researched and accurate. The claims check out. Recommend approval, with a few minor corrections and one design point worth probing.**

---

## 1. Protocol claims vs. the draft spec — accurate

Every spec claim I could verify is correct, including the subtle ones:

| Claim | Verdict |
|---|---|
| Removes sessions, `Mcp-Session-Id`, `initialize`, GET SSE stream, resumability | ✅ changelog major 1, 2, 4, 9 |
| Removes `ping`, `logging/setLevel`, `notifications/roots/list_changed`; server-initiated → MRTR | ✅ changes 5, 7 |
| `resultType` required; only `"complete"`/`"input_required"`; `| string` = extension point | ✅ schema `ResultType`, `Result.required=["resultType"]` |
| Error codes `-32020`/`-32021`/`-32022` (renumbered; SEP-2243 still cites `-32001`) | ✅ changelog minor 12 — exact match; the doc's caveat about the SEP is correct |
| `ttlMs`/`cacheScope` **required** on `CacheableResult` (no `?`) → omitting = schema-invalid | ✅ schema `required:["cacheScope","resultType","ttlMs"]`. The doc is *right*; the caching prose about "absent → default 0" is legacy-tolerance, not permission for modern servers |
| `server/discover` = capability **flags only**, no item arrays → topology leak, not enumeration | ✅ `DiscoverResult.capabilities = $ref ServerCapabilities`; `ServerCapabilities.properties=[completions,experimental,extensions,logging,prompts,resources,tools]` — flags, not items |
| `_meta`: `protocolVersion`+`clientCapabilities` required, `clientInfo` optional | ✅ schema `RequestMetaObject` |
| MRTR `requestState`: integrity **MUST**, principal/TTL/scope **SHOULD** | ✅ MRTR page server-reqs 4 & 5, verbatim |
| Era is "a property of the server, not of an individual request" | ✅ lifecycle Backward-Compat |
| Server-minted handles = SEP-2567 | ✅ changelog change 1 |
| agentgateway `stateless_send_and_initialize` prior art | ✅ real, role as described |

## 2. Codebase claims vs. `toolhive@main` — accurate (trivial line drift only)

The "cross-checked against origin/main" claim is justified:

- ✅ `coreVMCP` stateless + "core filters, Serve caches" — `core_vmcp.go:36-38`; `FilterTools` `:279`; `NewInMemoryStateStore` `:147`
- ✅ `var mcpHandler http.Handler = streamableServer` at `server.go:607`; `vmcpratelimit.NewDecorator` `:466`; `WriteTimeout` `:655`; `lazyInjectSessionTools` `:1064`
- ✅ Call-gate: `authzCallGate()` `call_gate.go:44`; `WithCallGate` gated on `authzGateEnabled` `server.go:543-544`
- ✅ `binding.Format(iss,sub)` → `iss + "\x00" + sub` verbatim (`binding.go:47`)
- ✅ Composer store keyed on `WorkflowID`; `WorkflowStatus` has **no owner field**; `ExecuteWorkflow` synchronous, `defer DeleteContext`, `WorkflowID` never persisted/returned → D5's "no migration needed" holds
- ✅ `pkg/audit` only whole-body switches + event exclusion — **no field-level redaction** (confirms "redaction is a build item")
- ✅ `defaultCacheCapacity = 1000`; both `.claude/rules/security.md` rules present
- ✅ Both backend `Initialize` calls fire with `LATEST_PROTOCOL_VERSION` (`client.go:677`, `mcp_session.go:415`)

## 3. Referenced issues/PRs — all exist; three minor mischaracterizations

Won't affect the technical argument, but worth fixing:

1. **PR #5839** — RFC says `ClassifyRevision`/`revision.go` was "added in #5839." It was **added in #5834** and *modified* by #5839.
2. **PR #5808** — design doc says "pending, not yet merged." It's actually **closed without merge** (2026-07-17), so the transport-proxy design doc isn't currently on a path to `docs/arch/` via that PR. "Pending" is misleading.
3. **Issue #5306** — cited as the basis for the `Format(iss,sub)` encoding. It's actually about **session churn on token refresh** (the binding breaking when a refreshed token changes). Related, but not *about* the encoding.

---

## Points worth probing before merge

**A. Kill-switch mechanism unspecified.** Defaulting Modern to off until "proven" is defensible (the asymmetry with RFC-0081 is well-argued). But the RFC says "without a redeploy" — implying a runtime toggle — without specifying the mechanism (config reload? dynamic flag?). Phase 1 is already safe-by-construction, so a flag that *also* defaults off means the feature ships dark and needs operator action to ever activate. One clarifying sentence on runtime-vs-startup would help.

**B. The one new attack surface is honestly flagged but thin.** The hand-rolled JSON-RPC envelope + re-homed call-gate is correctly identified as the main risk. "Lift the SDK's marshaling where possible, retire into v1.7" is the right direction, but the RFC is light on *how much* can actually be lifted from `mcpcompat` today before v1.7. Since `mcpcompat` lives in toolhive-core (Alpha), it's worth confirming whether the envelope marshaling is reusable as-is or needs extraction — a concrete pre-implementation question the RFC could answer.

**C. `server/discover` post-admission cost is understated.** Computing capability flags from the post-admission projection is the *right* security call. But "flag appears only when the identity is admitted to ≥1 item behind it" means discover now runs the same aggregation+admission fan-out as `tools/list` — on what the spec intends as a cheap, cacheable probe that clients MAY call before every session. Combined with the doc's own "Modern path pays a backend fan-out on every request" warning, discover risks becoming the *most frequent* expensive call. Caching is deferred to #5761 but not reconciled here. Not a blocker — worth an explicit note.

---

## Recommendation

**Approve.** The central thesis — vMCP's core is already stateless, so the work is *bypassing* the session bridge on the Modern path rather than rebuilding it — is verified true in the code and is the correct architectural read. Owning the off-diagonal bridge (D3) and not gating on go-sdk v1.7 (D2) are both sound. The security invariants (owner-bound handles, `(principal, backend)` partitioning, zero-credential pooling) are correctly designed and the spec's security requirements are quoted accurately.

Request the three reference corrections (#5839→#5834, #5808 closed-not-pending, #5306 reframing) and a clarifying sentence on (A) before merge. (B) and (C) can be tracked in #5756 rather than blocking.

JAORMX (dismissed):

JAORMX (approved):

